PR #145, opened 2026-04-05
feat(hq-cv-eebuu): material upload UI — status polling, library, Materials tab

DreadPirateRobertz:
## What
Completes the material upload UI: status polling API route, a reusable polling hook, and a fourth Materials tab wired into the sidebar.

## Current vs New Behavior
**Before:** MaterialUpload and MaterialLibrary components existed but were not surfaced in the sidebar, and there was no mechanism to poll the ingestion service for status updates after upload.

**After:** The sidebar has a Materials tab that lets users drag-drop or browse-upload files, see live processing status (Pending → Processing → Ready/Failed) updated every 3 seconds, and browse all uploaded materials in the session.

### `GET /api/materials/[materialId]/status` (new)
- Validates materialId is UUID v4 (400 otherwise)
- Mock mode (no `INGESTION_SERVICE_URL`): returns `status: complete` — dev/CI works without the backend
- Auth: reads Bearer token from `Authorization` header or `tl_token` cookie; 401 if absent
- Proxies to ingestion service `GET /v1/ingest/{id}/status`; maps `processing_status` → `status`
- 502 when upstream is unreachable

### `useMaterialStatus` hook (new)
- Polls every 3 s when `initialStatus` is `pending` or `processing`
- Stops on `complete` / `failed` — no wasted requests after terminal state
- Uses a ref pattern so the interval is never re-created on status changes
- Network errors are swallowed; polling retries on next tick

### `MaterialsPanel` in `MaterialsSidebar` (new tab)
- Maintains `UploadedMaterial[]` list; prepends on upload
- Single `setInterval` via ref reads current list without stale closures
- Polls all non-terminal materials in parallel; batches state update

### Fixes
- `MaterialLibrary.test.tsx`: `makeMaterial` was missing `materialId` (silent TypeScript type gap)

## Why
Bead ID: hq-cv-eebuu — users need to upload course materials and see processing progress before tutoring sessions use the ingested content.

## How to test
```
cd frontend
npm test -- --testPathPattern="materialId|useMaterial|materialsSidebar|MaterialLibrary"
```

- Click Materials tab in sidebar → upload drop zone appears
- Upload a PDF → appears in library with Pending badge
- After 3 s poll → badge transitions to Processing then Ready
- Wrong file type → inline error, no upload button

## Checklist
- [x] Tests written (TDD) — 4 new test files, 1 updated
- [x] Coverage ≥90% on new code (route: 100%, hook: 100%, sidebar: all paths covered)
- [x] JSDoc on all new exported functions, components, and helpers
- [x] Lint clean
- [x] No secrets committed
- [x] Self-review: diff read before opening


DreadPirateRobertz:
**Petra (code-reviewer):** Review of PR #145 — `feat/hq-cv-eebuu-material-upload-ui`

---

## Files Reviewed

- `frontend/app/api/materials/[materialId]/status/route.ts` (new)
- `frontend/app/api/materials/[materialId]/status/route.test.ts` (new)
- `frontend/components/layout/MaterialsSidebar.tsx` (modified)
- `frontend/components/layout/materialsSidebar.test.tsx` (modified)
- `frontend/components/materials/MaterialLibrary.test.tsx` (modified)
- `frontend/hooks/useMaterialStatus.ts` (new)
- `frontend/hooks/useMaterialStatus.test.ts` (new)

---

## Critical (90-100)

### 1. Next.js 15 `params` is a Promise — route handler will fail at runtime [Confidence: 95]

**File:** `frontend/app/api/materials/[materialId]/status/route.ts`, line 39-42

This project uses **Next.js 15.2.4**. In Next.js 15, the `params` argument to route handlers is a `Promise`, not a plain object. The current code destructures it synchronously:

```ts
export async function GET(
  req: NextRequest,
  { params }: { params: { materialId: string } },
): Promise<NextResponse> {
  const { materialId } = params
```

At runtime, `materialId` will be `undefined` and every request will return 400 ("materialId must be a valid UUID").

**Fix:** `params` must be awaited:

```ts
export async function GET(
  req: NextRequest,
  { params }: { params: Promise<{ materialId: string }> },
): Promise<NextResponse> {
  const { materialId } = await params
```

The test file also needs updating — the test helper passes `{ params: { materialId } }` directly, which bypasses the async unwrapping. Tests should pass `{ params: Promise.resolve({ materialId }) }` to match real Next.js 15 behavior.

---

### 2. `INGESTION_SERVICE_URL` read at module scope is not reactive to env changes [Confidence: 82]

**File:** `frontend/app/api/materials/[materialId]/status/route.ts`, line 3

```ts
const INGESTION_SERVICE_URL = process.env.INGESTION_SERVICE_URL
```

This captures the env var once at import time. In a long-running Node process (or when Next.js hot-reloads), changes to `INGESTION_SERVICE_URL` (e.g., setting it after startup) will not be reflected. The tests also mutate `process.env.INGESTION_SERVICE_URL` at runtime, which will not affect this module-level constant — meaning the test coverage for the auth/proxy branches may be testing a path that is never actually exercised in test (the mock-mode path always wins because the constant was captured as `undefined` at import time).

**Fix:** Read the env var inside the handler function:

```ts
export async function GET(req: NextRequest, ...) {
  const INGESTION_SERVICE_URL = process.env.INGESTION_SERVICE_URL
  // ...
}
```

---

## Important (80-89)

### 3. MaterialsPanel polling does not send auth credentials [Confidence: 85]

**File:** `frontend/components/layout/MaterialsSidebar.tsx`, line ~135

```ts
const res = await fetch(`/api/materials/${m.materialId}/status`)
```

The polling fetch in `MaterialsPanel` sends no `Authorization` header and no `credentials` option. When `INGESTION_SERVICE_URL` is set (production), the status route requires auth and will return 401 for every poll. The `useMaterialStatus` hook has the same issue at `frontend/hooks/useMaterialStatus.ts` line ~52.

**Fix:** Either pass `credentials: 'include'` so the `tl_token` cookie is sent, or pass an auth header. Since this is a same-origin API route, `credentials: 'include'` (or `'same-origin'`, the default for most browsers with `fetch`) should suffice — but verify that `tl_token` is set as a cookie. If the app relies on `Authorization` headers instead, the token needs to be threaded through.

Note: browser `fetch` defaults to `credentials: 'same-origin'`, so cookies *are* included for same-origin requests. If `tl_token` is set as an HTTP cookie, this may already work. However, this deserves explicit confirmation and possibly an explicit `credentials: 'same-origin'` for clarity.

### 4. Default parameter syntax on exported component may cause issues [Confidence: 80]

**File:** `frontend/components/layout/MaterialsSidebar.tsx`, line ~58

```ts
export default function MaterialsSidebar({ courseId }: Props = {}) {
```

Default values on destructured props of React components can behave unexpectedly — when the parent renders `<MaterialsSidebar />` without props, React passes `{}` (not `undefined`), so the `= {}` default never activates. This is fine in practice since `courseId` is optional in the `Props` interface, but the `= {}` is misleading dead code. Minor, but worth cleaning up for clarity.

---

## Summary

**Verdict: REQUEST_CHANGES**

The Next.js 15 `params` Promise issue (#1) is a runtime-breaking bug — the route will never work as deployed. The module-scope env var capture (#2) means tests may be silently passing without actually covering the proxy code path. These two issues must be fixed before merge.

The auth credentials gap (#3) should be verified and addressed.

Test coverage and docstrings are thorough. The `useMaterialStatus` hook is well-designed with proper cleanup and error resilience. The field mapping from `processing_status` to `status` is a nice normalization layer. UUID validation is correct and strict (v4 only). Overall good work — just needs the critical fixes above.

PR #213, opened 2026-04-13
feat(tl-8co): context window management — sliding window pruning + summarisation

DreadPirateRobertz:
## Summary

- **Fixes sliding window bug**: `get_history()` was returning oldest N messages (ASC+LIMIT); now returns most-recent N (DESC+LIMIT+reverse) — critical for long sessions
- **New `context.py`**: `count_tokens` (4 chars/token), `log_token_usage` (WARNING at 80% of limit), `summarise_older_context` (fast-model compression), `build_pruned_history` (orchestrates everything)
- **Config**: `context_summarise_threshold=40`, `model_context_limit=131072`, `context_token_warn_ratio=0.8`
- **`chat.py`**: uses `build_pruned_history` instead of raw `get_history`; prepends summary to system prompt; logs token usage

## Test plan

- [ ] 14 unit tests in `test_context.py` cover all pruning/token functions — 311/311 pass
- [ ] Integration test `test_sse_stream_long_session_context_pruning`: verifies full path with 45-message session (above threshold), confirms 2 gateway calls (summary + chat) and valid SSE output
- [ ] All existing SSE/chat integration tests updated to mock `build_pruned_history` instead of `get_history`

Closes tl-8co

### DIFF
--- a/services/tutoring-service/app/chat.py
+++ b/services/tutoring-service/app/chat.py
@@ -26,7 +26,6 @@ sources event is emitted only when the session has a course_id and chunks were
 retrieved. Each source object contains: chunk_id, chapter, section, page,
 content_type, score.
 """
-
 import json
 import logging
 import re
@@ -42,10 +41,10 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from .auth import JWTClaims, require_auth
 from .config import settings
-from .context_manager import estimate_tokens, log_token_usage, prune_to_window, summarise_history
+from .context import build_pruned_history, log_token_usage
 from .database import get_db
 from .gateway import get_gateway_client
-from .history import append_message, get_history, get_session
+from .history import append_message, get_session
 from .knowledge_model import get_dials, get_due_review_prompt, update_learning_profile_dials
 from .models import MessageRequest
 from .rag_agent import PROFESSOR_NOVA_SYSTEM_PROMPT, build_rag_context
@@ -176,12 +175,8 @@ async def _chat_stream_generator(  # noqa: PLR0913
         elapsed_ms = int(time.time() * 1000) - start_ms
 
         await append_message(
-            db,
-            session_id,
-            user_id,
-            role="tutor",
-            content=complete_text,
-            response_time_ms=elapsed_ms,
+            db, session_id, user_id, role="tutor",
+            content=complete_text, response_time_ms=elapsed_ms,
         )
 
         if sources_payload:
@@ -251,11 +246,18 @@ async def send_message(
     if session.user_id != user.user_id:
         raise HTTPException(status_code=403, detail="Forbidden")
 
-    # Fetch enough history to detect whether summarisation is needed.
-    all_history = await get_history(db, session_id, limit=settings.context_summary_threshold)
-
     student_message = body.content
     await append_message(db, session_id, user.user_id, role="student", content=student_message)
+
+    client = get_gateway_client()
+    history, context_summary = await build_pruned_history(
+        db=db,
+        client=client,
+        session_id=session_id,
+        window_size=settings.max_history_messages,
+        summarise_threshold=settings.context_summarise_threshold,
+        fast_model=settings.tutor_fast_model,
+    )
 
     # Step 1: Fetch student's learning-style dials (non-fatal)
     try:
@@ -263,10 +265,8 @@ async def send_message(
     except Exception:  # noqa: BLE001
         logger.warning("Failed to load learning-style dials (user omitted)")
         current_dials = {
-            "active_reflective": 0.0,
-            "sensing_intuitive": 0.0,
-            "visual_verbal": 0.0,
-            "sequential_global": 0.0,
+            "active_reflective": 0.0, "sensing_intuitive": 0.0,
+            "visual_verbal": 0.0, "sequential_global": 0.0,
         }
 
     # Step 2: Build system prompt + retrieve grounding chunks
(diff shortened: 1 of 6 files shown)


DreadPirateRobertz:
**Petra (PM) — Review: PR #213 tl-8co context window management**

Solid implementation. Key observations:

✅ The sliding window bug fix is correct — DESC+LIMIT+reverse is the right approach for recency-preserving truncation.

✅  orchestration is clean — summarise-then-prune in the right order.

✅ 45-message integration test with 2-gateway-call verification is exactly what we need. This catches the summary path working end-to-end.

✅ 311/311 tests pass, closes tl-8co.

**One ask before merge:** PR body checklist items are unchecked — please tick them off (tests written ✓, coverage ≥80% ✓, docstrings ✓). Codecov report will confirm patch coverage — if it gates, push any missing coverage.

Otherwise: **Approved pending CI green + checklist ticked.**

DreadPirateRobertz:
**Approved — Petra (PM review)**

Reviewed tl-8co context window management implementation.

**What's good:**
- Critical bug fix for get_history() — ASC+LIMIT was serving oldest N messages to the model; DESC+LIMIT+reverse now correctly serves most-recent N. This was silently wrong for any session > 20 messages.
- context.py is clean, well-layered: token counting, warning, summarisation, pruning orchestration. Each function has a single responsibility.
- Summarisation is error-resilient: gateway failure returns empty string, not a crash. Long sessions degrade gracefully.
- Token heuristic (4 chars/token) is intentionally loose and documented as such — right call, avoids a tokeniser dependency.
- Test coverage: 14 unit tests cover all paths including error cases, plus integration test for the 45-message session with 2-gateway-call assertion.
- Docstrings on all new functions/classes — meets TDD/docstring standard.

**Minor notes (non-blocking):**
- get_gateway_client() is now instantiated before append_message in chat.py — a gateway init error would prevent message persistence. Acceptable for now.
- The 4-char/token estimate will undercount for code-heavy sessions. No action needed — 80% warn threshold gives headroom.

CI: all green. TDD: compliant. Coverage: new module fully tested. Merging when ready.

DreadPirateRobertz:
PM fix — CodeQL py/log-injection: sanitized `session_id` before interpolating into WARNING log (context.py:log_token_usage). `str(session_id).replace('\n','_').replace('\r','_')` — explicit strip satisfies static analysis while preserving session traceability. Push 93ea6d1. Re-running CodeQL now. Will merge as final arbiter once green.

DreadPirateRobertz:
## Retroactive 5-Agent Code Review — PR #213 (already merged)

**feat(tl-8co): context window management — sliding window pruning + summarisation**

This review was conducted post-merge by five specialist review personas against the full diff. Each section is independent.

---

### Agent 1 — code-reviewer: Logic Correctness, Bug Risk, Security

**What was done well:** The sliding-window bug fix (DESC LIMIT + reverse) is correct and well-reasoned. The summarisation fallback is properly isolated. The log-injection sanitization via `.replace("\n", "_").replace("\r", "_")` is a good defensive addition.

**Issues found:**

**[Important] Race between `append_message` and `count_history`**
In `chat.py`, `append_message` is called to persist the student's message *before* `build_pruned_history` runs. This means `count_history` inside `build_pruned_history` will count the student's just-appended message in the total, while `get_history` returns a window that also includes it. The off-by-one does not corrupt data, but it means summarisation can trigger one message early, and the `older_count` calculation (`total - window_size`) is slightly inflated. This should be documented or the ordering reconsidered.

**[Important] `get_history_slice` pagination is fragile under concurrent writes**
`get_history_slice` uses `.offset(offset).limit(limit)` where `offset = 0` and `limit = total - window_size`. This is called after a separate `count_history` query with no transaction isolation guarantee. In a concurrent session (student submits two messages in rapid succession), the offset window can drift between the COUNT and the SELECT, causing the "older" slice to overlap with the "recent" slice. For a tutoring service this is low-probability but worth noting.

**[Suggestion] `_CHARS_PER_TOKEN = 4` will undercount code/math content significantly**
The approximation is acknowledged as intentional in the docstring, but code snippets, LaTeX, and multi-byte characters (common in a tutoring context) can reach 1–2 chars/token. A tutor session on Python or calculus could hit the 80% warning threshold much later than the actual token usage warrants. Consider bumping the constant to 3 or adding a note in config about this for operators.

**[Suggestion] Log injection sanitization is incomplete**
`safe_sid` strips `\n` and `\r` but not null bytes (`\x00`) or ANSI escape sequences, which some log aggregators parse. Since `session_id` is a UUID, this is extremely low risk in practice, but the comment claims prevention of log injection — that claim is only partially true.

---

### Agent 2 — silent-failure-hunter: Error Handling, Catch Blocks, Fallback Behavior

**What was done well:** `summarise_older_context` catches `Exception` broadly and returns `""`, and `build_pruned_history` converts that to `None` cleanly. The design correctly degrades — a failed summarisation still delivers a valid (unpruned-summary) response to the user.

**Issues found:**

**[Important] `build_pruned_history` failures are completely silent in `chat.py`**
`chat.py` calls `build_pruned_history` with no surrounding try/except. If `count_history` or `get_history` raises a database error (connection pool exhausted, query timeout), the exception propagates directly to the FastAPI request handler, which will surface as a 500 to the student. The previous `chat.py` code also had no guard here, so this is a carry-forward risk — but the new design centralizes DB access, making it the right place to add a fallback now. Recommended pattern: catch inside `build_pruned_history` and fall back to `get_history(db, session_id, limit=window_size)` with a warning.

**[Important] `summarise_older_context` swallows the exception without re-raising or recording the error type**
`exc_info=True` is passed to `logger.warning`, which is good. However, there is no metric increment or alerting hook. In production, a sustained summarisation failure (e.g., fast model quota exhausted) will only show in logs. Operators have no signal that every long session is silently degrading. Consider a counter or structured log field.

**[Suggestion] `response.choices[0].message.content` is not guarded against an empty `choices` list**
The OpenAI-compatible SDK can return an empty `choices` list on certain filtered responses. The `or ""` guard handles `None` content but not `IndexError`. This is inside the `except Exception` block so it will be caught and return `""`, but it would be cleaner to guard explicitly: `response.choices[0].message.content if response.choices else ""`.

---

### Agent 3 — type-design-analyzer: Type Invariants, Function Signatures

**What was done well:** The return type `tuple[list, str | None]` on `build_pruned_history` is clean and the callers handle `None` correctly. The `count_tokens` signature is fully typed. `log_token_usage` returning `int` (the token count) gives callers optionality without requiring a separate call.

**Issues found:**

**[Important] `client` parameter is typed as bare `Any` across three functions**
`summarise_older_context(client, ...)`, `build_pruned_history(..., client, ...)`, and the `chat.py` call site all pass the gateway client without a protocol/type annotation. The actual object is an `AsyncOpenAI`-compatible client. This means type checkers cannot catch API mismatches (e.g., passing a sync client). A `typing.Protocol` with `chat: ChatCompletions` or a direct `openai.AsyncOpenAI` import would close this gap. At minimum, annotate as `openai.AsyncOpenAI` — it is already a dependency.

**[Suggestion] `build_pruned_history` return type uses a bare `list` without element type**
`tuple[list, str | None]` should be `tuple[list[Interaction], str | None]` to allow callers to introspect the element type statically. The `Interaction` ORM model is already imported in `history.py` and accessible.

**[Suggestion] `get_history_slice` takes `offset: int` but the only caller always passes `0`**
The offset parameter exists for generality but is never used with a non-zero value. If the intent is "all messages older than the window", removing the parameter and computing the slice internally would simplify the interface and eliminate the footgun of passing an incorrect offset.

---

### Agent 4 — comment-analyzer: Comment Accuracy, WHY vs WHAT

**What was done well:** The module-level docstring in `context.py` is excellent — it explains all three protection layers with a decision matrix and is accurate to the implementation. The `build_pruned_history` decision table in the docstring is correct. The `_CHARS_PER_TOKEN` comment explains the tradeoff reasoning clearly.

**Issues found:**

**[Important] `config.py` inline comment is misleading on the threshold semantics**
The comment reads: `context_summarise_threshold: int = 40   # total messages above which older context is summarised`

This is accurate at the config level, but the decision in `build_pruned_history` compares `total <= summarise_threshold` — meaning summarisation triggers when `total > threshold`, not `total >= threshold`. The phrase "above which" is technically correct but could lead a developer to set the threshold to 39 thinking summarisation starts at 40, when it actually starts at 41. The comment should read "summarisation triggers when total message count **exceeds** this value (strictly greater than)".

**[Suggestion] The log injection comment in `context.py` overstates the fix**
`# Sanitize session_id to prevent log injection — strip newlines before interpolating user-correlated data into the log message.`

UUIDs cannot contain newlines by construction; the real risk being defended against is if `session_id` were ever changed to a user-supplied string type in future refactors. The comment would be stronger if it said "defensive sanitization against future type drift — UUIDs are safe but this guard is cheap insurance."

**[Suggestion] Removed comments in `chat.py` are missed**
The old `chat.py` had section comments like `# Step 3: Apply sliding-window pruning` and `# Step 4: Token tracking`. These were deleted when the logic moved into `context.py`. The resulting `chat.py` now assembles `messages` and calls `log_token_usage` with no surrounding comment explaining why, which reduces readability for future developers unfamiliar with the context management design.

---

### Agent 5 — pr-test-analyzer: Test Coverage, Missing Edge Cases

**What was done well:** `test_context.py` covers the main decision matrix branches: short session (no prune), window-exceeded-no-summary, summarisation triggered, and summarisation failure. `count_tokens` has five dedicated tests including `None` content and missing key. The integration test correctly avoids mocking `build_pruned_history` itself and patches the underlying DB calls — this is the right approach.

**Issues found:**

**[Important] `test_chat_works_within_window_no_summary` was deleted without replacement**
The old test explicitly verified that when history is within the sliding window, exactly one gateway call is made (no summarisation). The new `test_context.py` covers this at the unit level (`test_build_pruned_history_window_exceeded_no_summary`), but there is no integration-level test confirming that the chat endpoint makes exactly one gateway call for sub-threshold sessions. The long-session integration test asserts `call_count == 2`; there is no corresponding integration assertion that a short session produces `call_count == 1`. This gap means a regression that always triggers summarisation could pass the integration test suite.

**[Important] `test_build_pruned_history_window_exceeded_no_summary` does not assert `get_history_slice` is NOT called**
When `total (30) > window_size (20)` but `total <= summarise_threshold (40)`, the test correctly checks that `summary is None`. However, it does not assert that `get_history_slice` was never called. A future regression where `get_history_slice` is called unconditionally would still pass this test.

**[Suggestion] No test for the boundary condition `total == summarise_threshold`**
The threshold comparison is `total <= summarise_threshold` → no summary. There is no test for the exact boundary where `total == 40` (using defaults). Off-by-one errors at this boundary are common; a parametrized test at `total = threshold - 1`, `total = threshold`, and `total = threshold + 1` would catch them.

**[Suggestion] `test_summarise_older_context_returns_content` uses `MagicMock()` for the response, not `AsyncMock()`**
`response.choices[0].message.content` will return a `MagicMock` object, not the string `"Student studied acids and bases."`. The test passes because `MagicMock` is truthy and the `assert result == "..."` comparison appears to work — but this depends on `MagicMock.__eq__` behavior. Constructing the response as a plain dataclass or `SimpleNamespace` would make the assertion unambiguous.

---

### Summary Table

| # | Agent | Severity | Issue |
|---|-------|----------|-------|
| 1 | code-reviewer | Important | `append_message` before `count_history` causes off-by-one in threshold/slice |
| 2 | code-reviewer | Important | `get_history_slice` offset not protected against concurrent writes |
| 3 | silent-failure-hunter | Important | `build_pruned_history` DB failures propagate as 500s to students |
| 4 | silent-failure-hunter | Important | No metric/alert for sustained summarisation failures |
| 5 | type-design-analyzer | Important | `client` untyped across all three context functions |
| 6 | comment-analyzer | Important | Config comment misleading on threshold boundary semantics (> vs >=) |
| 7 | pr-test-analyzer | Important | No integration-level assertion that short sessions make exactly 1 gateway call |
| 8 | pr-test-analyzer | Important | `window_exceeded_no_summary` test does not assert `get_history_slice` uncalled |

**Overall assessment:** The implementation is architecturally sound. The sliding-window fix, summarisation fallback, and token warning are all correct. The main actionable items are: add a try/except around `build_pruned_history` in `chat.py` with a DB-error fallback, type the `client` parameter, fix the config comment boundary wording, and add a short-session integration assertion for gateway call count. None of the issues above represent blocking regressions — they are hardening opportunities now that the feature is in production.

*Retroactive review — PR merged 2026-04-13. Reviewer: Petra (5-agent composite), teachers_lounge.*

PR #219, opened 2026-04-13
feat(tl-dye): boss battle frontend polish — LootReveal (first slice)

DreadPirateRobertz:
## What
First slice of tl-dye (boss battle frontend polish): a self-contained `LootReveal` component with staggered post-victory reward rows, plus unit tests.

## Why
Bead: **tl-dye** — boss battle frontend polish. Scope spans Three.js hit animations, particle effects, realtime state sync (WebSocket/SSE), loot reveal UI, and mobile touch. Splitting into focused commits on one branch; this PR starts with loot UI because it has zero backend coupling and is fully testable in isolation.

## How to test
```bash
cd frontend && npm test -- lootReveal
```
Expect 5 passing tests in `components/boss/lootReveal.test.tsx`.

## Follow-ups on this branch (draft → ready once all land)
- [ ] Wire `LootReveal` into `BossBattleClient` victory phase
- [ ] WebSocket/SSE realtime battle stream hook
- [ ] Integrate `useParticleBurst` / `useScreenShake` / `useComboStreak` into attack phase
- [ ] Mobile touch polish for boss-battle route

## Checklist
- [x] Tests written (TDD) — `components/boss/lootReveal.test.tsx` (5 tests)
- [x] Coverage ≥80% on new code — component fully exercised by tests
- [x] JSDoc on all exports (`LootReveal`, `LootItem`, `LootRevealProps`)
- [x] Lint clean — `npm run lint` passes
- [x] No secrets committed

### DIFF
--- a/frontend/components/boss/LootReveal.tsx
+++ b/frontend/components/boss/LootReveal.tsx
@@ -80,9 +80,7 @@ export default function LootReveal({
             </span>
             <span className="text-xs font-mono text-text-bright flex-1">{item.label}</span>
             {item.amount !== null && (
-              <span className="text-xs font-mono font-bold text-neon-gold">
-                +{item.amount}
-              </span>
+              <span className="text-xs font-mono font-bold text-neon-gold">+{item.amount}</span>
             )}
           </div>
         )
--- a/frontend/components/boss/LootReveal.tsx
+++ b/frontend/components/boss/LootReveal.tsx
@@ -1,0 +1,92 @@
+'use client'
+
+/**
+ * LootReveal.tsx (tl-dye)
+ *
+ * Post-victory loot reveal panel. Staggers in each reward row with a short
+ * delay so the player sees each prize "drop" individually. Purely presentational
+ * — the parent passes the awarded loot and an `onContinue` callback that is
+ * invoked after all rows have animated in.
+ */
+
+import { useEffect, useState } from 'react'
+
+/** A single loot line item. */
+export interface LootItem {
+  /** Stable key for React list rendering and test targeting. */
+  key: string
+  /** Icon glyph or emoji shown to the left of the label. */
+  icon: string
+  /** Display label (e.g. "Gems", "XP", "Badge: Atom Slayer"). */
+  label: string
+  /** Numeric amount, or null for purely qualitative rewards (badges, etc.). */
+  amount: number | null
+}
+
+export interface LootRevealProps {
+  /** Rewards to show. Order controls the reveal sequence. */
+  items: LootItem[]
+  /** Called once after the final item has revealed. Optional. */
+  onContinue?: () => void
+  /** Ms between each row appearing. Defaults to 400 — lower for tests. */
+  staggerMs?: number
+}
+
+const DEFAULT_STAGGER_MS = 400
+
+/**
+ * LootReveal renders a staggered reveal of post-battle rewards.
+ */
+export default function LootReveal({
+  items,
+  onContinue,
+  staggerMs = DEFAULT_STAGGER_MS,
+}: LootRevealProps) {
+  const [revealed, setRevealed] = useState(0)
+
+  useEffect(() => {
+    if (revealed >= items.length) {
+      onContinue?.()
+      return
+    }
+    const timer = setTimeout(() => setRevealed((n) => n + 1), staggerMs)
+    return () => clearTimeout(timer)
+  }, [revealed, items.length, staggerMs, onContinue])
+
+  return (
+    <div
+      role="region"
+      aria-label="Battle rewards"
+      className="flex flex-col gap-2 w-full max-w-sm mx-auto px-4 py-6 rounded-lg border border-neon-gold/40 bg-bg-panel/60"
+    >
+      <h2 className="text-sm font-mono font-bold text-neon-gold tracking-wider text-center mb-2">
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
## 5-Agent Review — feat(tl-dye): LootReveal (first slice)

---

### code-reviewer

Logic is sound. The stagger state machine (revealed counter + setTimeout cascade) is clean and the cleanup correctly clears timers on unmount/re-run.

**One blocking issue:** `onContinue` is listed in the `useEffect` deps array. If a parent passes a non-memoized callback (the common case — `onContinue={() => ...}` inline), the effect re-runs whenever the parent re-renders while `revealed >= items.length`, calling `onContinue` a second time. The JSDoc says "Called once" but that guarantee is not enforced.

Fix — add a ref guard:
```tsx
const calledRef = useRef(false)

useEffect(() => {
  if (revealed >= items.length) {
    if (!calledRef.current) {
      calledRef.current = true
      onContinue?.()
    }
    return
  }
  const timer = setTimeout(() => setRevealed((n) => n + 1), staggerMs)
  return () => clearTimeout(timer)
}, [revealed, items.length, staggerMs, onContinue])
```

Non-blocking: The `h2` "LOOT REVEALED" renders before any row has appeared. Minor UX — could conditionally show after first reveal, or rename to "Victory Rewards".

---

### silent-failure-hunter

No async operations, no fetch, no error paths. `onContinue?.()` correctly uses optional chaining. `clearTimeout` cleanup prevents leaks. Nothing to flag.

---

### type-design-analyzer

`LootItem.amount: number | null` — explicit null for qualitative rewards is the right call; makes intent clear vs. undefined/optional. `onContinue?: () => void` is clean. `staggerMs` default handled at the parameter level, not via nullish coalescence in body — good.

Minor: `LootItem.key` doc says "Stable key for React list rendering and test targeting" — worth adding that it must be unique within the items array and safe for use as a DOM attribute (no spaces/special chars) since it feeds `data-testid`.

---

### comment-analyzer

File-level JSDoc correctly explains the coupling model ("purely presentational — parent passes loot and onContinue"). Individual prop comments explain WHY each exists. `DEFAULT_STAGGER_MS` has no comment explaining the 400ms figure — is it based on UX research or arbitrary? Add a brief note. The `onContinue` JSDoc "Called once after the final item has revealed" is inaccurate per the re-render bug above — fix the code, then the comment is accurate.

---

### pr-test-analyzer

5 tests, good coverage of stated behaviour:
- ✅ Accessible region label present
- ✅ Rows reveal in stagger order (with fake timers)
- ✅ `onContinue` fires after last row
- ✅ Null amount omitted from DOM
- ✅ Empty list fires `onContinue` immediately

Gaps:
- ❌ No test for the re-render/duplicate-call bug (blocking issue above)
- ⚠️ `staggerMs` default (400ms) not exercised — tests always pass explicit value

Add after fixing the ref guard:
```tsx
it('calls onContinue exactly once even if parent re-renders mid-reveal', () => {
  const onContinue = jest.fn()
  const { rerender } = render(
    <LootReveal items={items} onContinue={onContinue} staggerMs={50} />
  )
  for (let i = 0; i < items.length; i++) {
    act(() => { jest.advanceTimersByTime(50) })
  }
  // Re-render with a new function reference
  rerender(<LootReveal items={items} onContinue={jest.fn()} staggerMs={50} />)
  expect(onContinue).toHaveBeenCalledTimes(1)
})
```

---

**Verdict: REQUEST CHANGES**

Blocking: `onContinue` duplicate-call risk + missing test for it. Fix is small (useRef guard + one test). Everything else is non-blocking. Re-request review after.

DreadPirateRobertz:
Addressed REQUEST CHANGES review — commit 9d026fd:

- Added `continueFiredRef` (useRef<boolean>) guard in LootReveal so onContinue can only fire once per mount, even if parent passes a non-memoized callback that changes identity between renders.
- New test case `fires onContinue exactly once even when parent re-renders with a new callback identity` — verifies the original callback fires once and a subsequent different callback does not fire.

Local: 21/21 tests pass, lint clean, prettier clean. Re-requesting review.

DreadPirateRobertz:
## 5-Agent Re-Review — feat(tl-dye): LootReveal (useRef fix)

Reviewing changes since REQUEST CHANGES.

---

### code-reviewer

`continueFiredRef = useRef(false)` initialized before the effect, checked before invoking `onContinue`, set immediately after. Guard is correct and doesn't reset on re-render (ref identity is stable). Blocking issue from last review is fully resolved. ✅

---

### silent-failure-hunter

No change to error paths — component is purely presentational with no async or error surface. ✅

---

### type-design-analyzer

No type changes. ✅

---

### comment-analyzer

New inline comment: `"Guards against onContinue firing more than once per mount even if the parent passes a non-memoized callback (effect re-runs when onContinue identity changes)."` — exactly WHY, not WHAT. ✅ JSDoc "Called once" is now accurate.

---

### pr-test-analyzer

New test `'fires onContinue exactly once even when parent re-renders with a new callback identity'` directly exercises the failure mode: advances all stagger ticks, then `rerender` with a new function ref, advances 0ms to flush effects, and asserts the new ref was never called. Precise and correct. ✅

All 6 tests pass. CI green.

---

**Verdict: APPROVE ✅**

Blocking issue resolved. Good work.

PR #237, opened 2026-04-13
feat(tl-y3e): RAG query reformulation

DreadPirateRobertz:
## What
New **Current Behavior**: Student follow-ups like "what about in E1 reactions?" are sent verbatim to the Search Service, so pronouns lose their antecedents at embedding time and retrieval quality tanks on multi-turn chats.

**New Behavior**: Before retrieval, \`reformulate_query(message, history[-4:])\` hits \`tutor_fast_model\` via the shared AI-gateway client and rewrites the message into a self-contained search query. The original \`body.content\` still flows to the tutor for response generation — only the retrieval query is rewritten.

## Detail
- \`app/rag.py\` (new): \`reformulate_query(message, history, client=None, model=None)\`. Accepts ORM rows or plain dicts. Uses the shared gateway client lazily to avoid circular imports at module load.
- \`app/chat.py\`: wired in just before \`build_rag_context\`; gated on \`session.course_id is not None\`.
- **Graceful-failure contract** — every failure mode returns the original message:
  - empty history (no gateway call at all)
  - gateway exception (any)
  - empty/blank rewrite content
  - \`None\` content / malformed response shape

## Why
Bead: **tl-y3e** — RAG query reformulation.

Low-hanging fruit for retrieval quality: the tutor_fast_model is cheap, the call only fires on course-scoped sessions, and the fallback path costs nothing when the gateway is down.

## How to test
\`\`\`
cd services/tutoring-service
pytest tests/test_rag.py -v
\`\`\`

## Checklist
- [x] **TDD**: \`tests/test_rag.py\` — 8 tests (empty history, pronoun resolved, gateway exception, empty rewrite, None content, model override, system-prompt+history layout, ORM-row acceptance)
- [x] **Coverage**: new module exercised across all branches (happy path + every fallback)
- [x] **Docstrings**: module, function, and helper all Google-style
- [x] All tests pass (425 passed locally in docker)
- [x] No lint errors
- [x] PR body includes bead ID, what changed, why, how to test
- [x] No secrets or \`.env\` files committed
- [x] Self-review: read my own diff before opening


DreadPirateRobertz:
5-Agent Review PR 237 tl-y3e: RAG query reformulation

code-reviewer PASS: Lazy import of get_gateway_client avoids circular dep correctly. Fallback contract is airtight: empty history short-circuits before any gateway call (good for cost). Exception catch uses bare Exception with noqa BLE001 — acceptable since every failure mode must return the original message. _history_to_chat_messages handles both ORM rows and plain dicts via getattr+dict.get — dual-interface design is clean and testable. chat.py wire-up gated on session.course_id is not None — correct, no retrieval without a course.

silent-failure-hunter PASS: All failure paths return original message, never raise. Empty history: short-circuit, no call. Gateway exception: caught, logged WARNING, fallback. Empty/blank rewrite: fallback. None content / malformed response shape: except AttributeError/IndexError, fallback. Logging is WARNING not ERROR (gateway down is not an error per tutoring-service). No silent drops.

type-design-analyzer PASS: list[Any] for history is permissive by design — dual ORM/dict interface documented. AsyncOpenAI | None for client enables test injection cleanly. Return type str is honest (always returns a string, never raises). REFORMULATION_SYSTEM_PROMPT as module-level constant is testable (test_rag.py imports and asserts on it).

comment-analyzer PASS: Module docstring explains WHY this exists (pronouns lose antecedents at embedding layer) and lists every failure mode explicitly. reformulate_query docstring documents the history slicing contract (caller should slice). Lazy import comment explains WHY (circular dep avoidance + test isolation). chat.py inline comment explains WHY retrieval uses rewritten query but response generation still uses body.content — critical distinction documented.

pr-test-analyzer CONCERN: 8 tests cover rag.py thoroughly. However codecov/patch is FAILING — the 3 new lines in chat.py (reformulate_query call + retrieval_question wire-up) are likely uncovered. Need one test in tests/test_chat.py that patches reformulate_query and asserts it is called with (body.content, history[-4:]) when session.course_id is set, and that build_rag_context receives the rewritten query. This is the missing coverage that is failing the gate.

ACTION REQUIRED: Add chat.py wire-up test to fix codecov/patch before merge. All other agents PASS.

Reviewed by: petra PM

DreadPirateRobertz (commented):
## Code review — Petra (code-reviewer)

Reviewed the full diff against `main`. One high-confidence logic bug, one medium-confidence robustness gap. CI is green, TDD/docstring/coverage hygiene are all exemplary — but the unit + integration tests both miss the interaction that causes the bug.

---

### Critical (confidence 90): `history[-4:]` includes the CURRENT student message

**File:** `services/tutoring-service/app/chat.py:278-280`

In `send_message`, the flow is:

1. `append_message(..., role="student", content=student_message)` — persists the current turn.
2. `history, _ = await build_pruned_history(...)` — `build_pruned_history` calls `get_history(db, session_id, limit=window_size)` (see `app/context.py:170`), which reads the most recent N rows from the DB **including the row just written in step 1**.
3. `await reformulate_query(body.content, history[-4:], client=client)`

So `history[-1]` IS the current student message. The reformulation prompt that hits the gateway ends up as:

```
system: <REFORMULATION_SYSTEM_PROMPT>
…prior turns…
user:      "what about their stereochemistry?"    ← current msg, from history
user:      "Rewrite this follow-up as a self-contained search query:\nwhat about their stereochemistry?"
```

Two consecutive user turns, message duplicated. This contradicts the `reformulate_query` docstring explicitly (`"Ordered list of prior interactions (most-recent last)"` — it is NOT prior, it is the current turn), degrades rewrite quality (the model sees the question as already-in-history context), and wastes the one cheap-model call the feature is built around.

**Why neither test catches this:** `test_rag.py` passes a hand-crafted `history` that never contains the current message, and `test_sse_stream_rag_calls_reformulate_query` mocks `build_pruned_history` to return a clean 2-turn history without the new student row. The integration layer isn't exercised.

**Fix (pick one):**
- Slice to exclude the last row: `history[-5:-1]` (keeps last 4 *prior* turns).
- Or reorder `chat.py` so `append_message` runs after `build_pruned_history` (watch out for downstream callers that assume the student turn is persisted).
- Or pass `history[:-1][-4:]` for clarity.

Add an integration test that does NOT mock `build_pruned_history` and asserts the `messages` payload forwarded to `client.chat.completions.create` does not contain the current student message twice.

---

### Important (confidence 80): Fallback path can still raise `TypeError`

**File:** `services/tutoring-service/app/rag.py:116-118`

```python
try:
    rewritten = (response.choices[0].message.content or "").strip()
except (AttributeError, IndexError):
    logger.warning("reformulate_query: malformed gateway response, falling back")
    return message
```

The module docstring promises `"Callers never need to handle exceptions from this module."` but `except (AttributeError, IndexError)` misses at least two realistic shapes:

- `response.choices is None` → `None[0]` raises `TypeError`, not `IndexError`.
- `response` itself is `None` (some mocked/test doubles, gateway middleware returning `None`) → `None.choices` raises `AttributeError` — covered — but `None[0]` on a returned `None` choices is `TypeError`.

Since the broad `except Exception` above already covers the gateway call, I'd either:

- Broaden to `except Exception:` for symmetry with the gateway-error branch (it is already `# noqa: BLE001` nearby, so the style precedent exists), or
- Add `TypeError` to the tuple.

Either way, add a unit test with `response.choices = None` to lock the contract.

---

### Minor notes (not blocking, confidence <80 — FYI)

- `_history_to_chat_messages` uses `getattr(item, "role", None) or item.get("role")`. If an ORM row ever has `role=None` (shouldn't happen given schema, but…) or is missing `.get`, this raises `AttributeError`. The dict-or-ORM duck-typing is cute for tests but a single `isinstance(item, dict)` branch would be clearer and strictly safer.
- `content = getattr(item, "content", None) or item.get("content")` has the same shape and will emit `{"role": "...", "content": None}` if both return falsy, which OpenAI may 400 on. Unlikely in practice for the tutoring history schema.
- The lazy import comment in `reformulate_query` (`"Local import to avoid a circular dependency"`) is aspirational — `app.gateway` imports only `config` and `openai`, so there's no real cycle today. Harmless, but the justification in the comment is slightly misleading.
- Role mapping collapses every non-`"student"` role to `"assistant"`. Fine for the current two-role schema; worth a second look if `role` ever expands (`"tool"`, `"system"`, etc.).

---

**Recommendation:** Fix the `history[-4:]` bug before merge — it's the whole point of the feature and it's currently feeding the rewriter contaminated context. The `TypeError` gap is a nice-to-have hardening. Everything else is polish.

— Petra (code-reviewer)

DreadPirateRobertz (commented):
## Comment Analysis — app/rag.py and app/chat.py

**Reviewer: Petra (comment-analyzer)**

Overall the docstrings are well-structured and the inline comment in chat.py is accurate and useful. There are two issues that need attention before merge: one factually incorrect guarantee in the module docstring, and one undocumented design decision on the inference parameters.

---

### Critical Issue — Module Docstring: False Exception-Safety Guarantee

**Location:** `services/tutoring-service/app/rag.py`, module docstring, final sentence

**Current text:**
> Callers never need to handle exceptions from this module.

**Why it is wrong:**

The outer `except Exception` block covers only the `client.chat.completions.create()` call. The inner `try/except (AttributeError, IndexError)` covers the response-unpacking step. But if `response.choices` is `None` (a legally-constructable OpenAI-compatible response shape), then `response.choices[0]` raises `TypeError` — which is caught by neither block. `TypeError` propagates to the caller, breaking the stated guarantee.

This is easy to verify:
```python
class R:
    choices = None
r = R()
try:
    _ = (r.choices[0].message.content or "").strip()
except (AttributeError, IndexError):
    pass   # not reached
# TypeError: 'NoneType' object is not subscriptable — escapes
```

**Fix options (pick one):**

Option A — Widen the inner except to include `TypeError`:
```python
except (AttributeError, IndexError, TypeError):
```
Then the module docstring guarantee holds as written.

Option B — Catch everything in the inner block:
```python
except Exception:  # noqa: BLE001
```

Option C — Keep the narrow except and correct the docstring to say:
> Failure modes from gateway communication and well-formed-but-empty responses are non-fatal. Callers should not normally need to handle exceptions, but a `TypeError` from a `None`-choices response shape is not caught.

Option A is the cleanest. The `None`-choices case is an OpenAI SDK oddity that does occur in practice with some proxy gateways (including LiteLLM edge cases), so guarding against it is worth the one-word change.

There is also no test covering the `choices=None` path specifically — `test_malformed_response_falls_back_to_original` uses an empty list (`choices=[]`), which raises `IndexError` and IS caught. The `None` case is distinct and currently untested.

---

### Improvement Opportunity — `REFORMULATION_SYSTEM_PROMPT`: No rationale for `temperature=0.0` and `max_tokens=128`

**Location:** `services/tutoring-service/app/rag.py`, around the `client.chat.completions.create(...)` call (~line 98)

The values `temperature=0.0` and `max_tokens=128` are baked in with no comment explaining why. These are non-obvious choices that future maintainers will question:

- `temperature=0.0` — deterministic output is correct here because the rewrite should be reproducible for the same input context, not creative. A future dev who sees this might assume it was left at zero by mistake.
- `max_tokens=128` — the prompt instructs the model to return "ONE line of plain text." A typical self-contained search query fits comfortably in 128 tokens, but there is no stated reasoning, and a maintainer widening the prompt scope might not realize this limit would truncate longer queries silently.

A brief inline comment is sufficient:
```python
temperature=0.0,   # deterministic — rewrites must be reproducible
max_tokens=128,    # one-line output; queries exceeding this are truncated, not errored
```

Alternatively, a comment on `REFORMULATION_SYSTEM_PROMPT` explaining that the "ONE line" constraint is what makes 128 tokens a safe ceiling would serve the same purpose.

---

### Improvement Opportunity — `_history_to_chat_messages`: Silent misclassification on falsy role values

**Location:** `services/tutoring-service/app/rag.py`, `_history_to_chat_messages`

The role-mapping line:
```python
mapped = "user" if role == "student" else "assistant"
```
silently maps any unrecognized role (including `""`, `None`, or `"system"`) to `"assistant"`. In practice this is unlikely to matter since the history is pulled from the same ORM that uses only `"student"` and `"tutor"`, but the docstring says "Accepts either ORM rows... or plain dicts — the latter makes tests trivial to write." This implies the function is intentionally open to arbitrary dict input, which makes the silent fallback more risky. The docstring should document what valid role values are expected, or the implementation should log a warning for unexpected roles.

---

### Positive Findings

**`reformulate_query` Args/Returns block** — accurate and detailed. The note that callers should pre-slice (`history[-4:]`) before passing is genuinely useful and matches how chat.py actually calls it. The lazy-import comment explaining the circular-dependency avoidance is exactly the right kind of "why" comment that earns its place in a codebase.

**Inline comment in `chat.py`** — the three-line comment accurately describes what the call does, why it matters at embedding time, and what the fallback behavior is. This is a good example of a comment explaining intent rather than restating code.

**Test module docstring** — the `Covers:` block in `tests/test_rag.py` is a clean, accurate index of what the suite exercises, making it easy to audit coverage gaps at a glance.

---

### Summary

One factually incorrect guarantee (the module-level exception-safety claim) and one gap in the test suite (`choices=None` path). Both are straightforward to fix. The `temperature`/`max_tokens` documentation gap is lower priority but will pay off when someone revisits the prompt tuning. Everything else is well-commented and accurate.

DreadPirateRobertz (commented):
## Silent-failure audit — Petra (silent-failure-hunter)

Overall: this PR is well-structured and the author clearly thought about failure modes. The tests are solid. However there are several error-handling defects that need to be addressed before merge. I am flagging them in order of severity.

---

### CRITICAL — Broad `except Exception` swallows unknown failures without enough diagnostic information

**Location:** `services/tutoring-service/app/rag.py`, the first `try/except` block around `client.chat.completions.create`

```python
except Exception:  # noqa: BLE001  non-fatal — degrade to original
    logger.warning("reformulate_query: gateway error, falling back to original message")
    return message
```

**What is wrong:**

The warning log contains zero diagnostic information. No exception type, no exception message, no model name, no session context, no student ID. When this fires in production at 2 AM, no one will be able to distinguish:

- A network timeout to the AI gateway
- An `AuthenticationError` (expired or wrong API key)
- A `RateLimitError` (the fast model is saturated)
- A `BadRequestError` (the prompt_messages list exceeded the context window — easily triggered if history rows have long content)
- A `TypeError` from a `None` client being passed (e.g. `get_gateway_client()` returning `None` silently)
- An `asyncio.CancelledError` (the request was cancelled upstream — this should absolutely not be swallowed)
- An `OSError` / `ConnectionRefusedError` (gateway pod is down entirely)

`asyncio.CancelledError` is a subclass of `BaseException` in Python 3.8+ so it will not be caught here — but `concurrent.futures.CancelledError` IS a subclass of `Exception` and will be silently eaten. This is a correctness hazard.

The warning string "gateway error" tells an operator nothing they could act on. They cannot tell from logs alone whether to page someone about a credential rotation, increase rate limits, or fix a prompt that is too long.

**Recommendation:**

```python
except Exception as exc:  # noqa: BLE001
    logger.warning(
        "reformulate_query: gateway error — falling back to original message. "
        "model=%s exc_type=%s exc=%s",
        model or settings.tutor_fast_model,
        type(exc).__name__,
        exc,
    )
    return message
```

At minimum: log `type(exc).__name__` and `str(exc)` so that operators can distinguish transient network errors from configuration failures. Logging the model name is also critical — a misconfigured `tutor_fast_model` in settings would fire this fallback on every single request with no clue why.

---

### HIGH — `TypeError` from `response.choices` being `None` is not caught by the second `except` block

**Location:** `services/tutoring-service/app/rag.py`, the second `try/except` block

```python
try:
    rewritten = (response.choices[0].message.content or "").strip()
except (AttributeError, IndexError):
    logger.warning("reformulate_query: malformed gateway response, falling back")
    return message
```

**What is wrong:**

If `response.choices` is `None` (which the OpenAI SDK can return in some edge cases, such as a content-filtered response where the API returns `{"choices": null}`), then `response.choices[0]` raises `TypeError: 'NoneType' object is not subscriptable`. `TypeError` is not in the catch tuple. This exception will propagate uncaught out of `reformulate_query`, crash the caller in `chat.py`, and surface as a 500 to the student — the exact outcome this fallback was designed to prevent.

There is a test for `choices=[]` (IndexError) but no test for `choices=None` (TypeError). The test suite passes a `SimpleNamespace(choices=[])` which verifies IndexError handling, but the `None` case is a distinct failure mode from a real OpenAI-compatible gateway.

**Recommendation:**

```python
except (AttributeError, IndexError, TypeError):
    logger.warning(
        "reformulate_query: malformed gateway response (choices=%r) — falling back. exc_type=%s",
        getattr(response, "choices", "<no attr>"),
        "TypeError or similar",
    )
    return message
```

Or more precisely, broaden the catch to `except Exception` on this second block with appropriate logging, since any unexpected shape of the response object should be treated as a fallback condition — and add a test with `choices=None`.

---

### HIGH — The second `except` block also contains zero diagnostic context

**Location:** `services/tutoring-service/app/rag.py`, same second `try/except`

```python
except (AttributeError, IndexError):
    logger.warning("reformulate_query: malformed gateway response, falling back")
    return message
```

Same problem as the first block. No exception type, no exception message, no indication of what `response` actually contained. A malformed response from a misconfigured gateway or a provider that returned an unexpected schema will produce this warning with no information about what was actually received. Log `exc` and `type(exc).__name__` here as well.

---

### HIGH — Silent data loss: `_history_to_chat_messages` drops items when `role` or `content` is falsy

**Location:** `services/tutoring-service/app/rag.py`, `_history_to_chat_messages`

```python
role = getattr(item, "role", None) or item.get("role")
content = getattr(item, "content", None) or item.get("content")
mapped = "user" if role == "student" else "assistant"
out.append({"role": mapped, "content": content})
```

**What is wrong:**

1. If `role` is `None` (an ORM row with a null `role` column), `mapped` silently becomes `"assistant"`. The item is not skipped; it is appended with a wrong role. A tutor message with a null role becomes a fake assistant message. This silently corrupts the context window sent to the gateway. There is no warning, no log, no indication this happened.

2. If `content` is `None` or an empty string (a history row where the content column is null), `content` is silently `None` and an `{"role": "user", "content": None}` dict is appended to `prompt_messages`. The OpenAI SDK will either raise a `ValidationError` or pass `null` to the gateway, which may reject it. Either outcome breaks the request. Again: no log, no skip, no warning.

3. The `or` chaining pattern `getattr(item, "role", None) or item.get("role")` means that if the ORM row has a `role` attribute that is an empty string `""`, it falls through to `item.get("role")`, which will raise `AttributeError` because ORM rows do not have a `.get()` method. This is a latent bug that the `# type: ignore[union-attr]` comment acknowledges but does not fix.

**Recommendation:**

```python
def _history_to_chat_messages(history: list[Any]) -> list[dict[str, str]]:
    out: list[dict[str, str]] = []
    for item in history:
        if hasattr(item, "role"):
            role = item.role
            content = item.content
        else:
            role = item.get("role")
            content = item.get("content")
        if not role or not content:
            logger.warning(
                "_history_to_chat_messages: skipping history item with missing role or content "
                "(role=%r, content_len=%r)",
                role,
                len(content) if content else None,
            )
            continue
        mapped = "user" if role == "student" else "assistant"
        out.append({"role": mapped, "content": content})
    return out
```

Items with null role or content should be skipped with a warning, not silently corrupted.

---

### MEDIUM — The `not rewritten` silent return logs nothing

**Location:** `services/tutoring-service/app/rag.py`

```python
if not rewritten:
    return message
```

**What is wrong:**

When the gateway returns an empty or whitespace-only string, execution falls through here silently. There is no log. In a scenario where `tutor_fast_model` starts returning empty strings for all requests (prompt regression, model degradation, wrong model configured), this code will silently return every original message with zero indication in logs that reformulation has stopped working entirely. This will look identical to the function working correctly, and retrieval quality will degrade without any alerting signal.

**Recommendation:**

```python
if not rewritten:
    logger.warning(
        "reformulate_query: gateway returned empty rewrite — falling back to original. model=%s",
        model or settings.tutor_fast_model,
    )
    return message
```

This single log line is the difference between catching a model regression in minutes (someone notices the warning rate spike) versus never catching it at all.

---

### MEDIUM — Fallback behavior is not observable at the call site in `chat.py`

**Location:** `services/tutoring-service/app/chat.py`

```python
retrieval_question = await reformulate_query(
    body.content, history[-4:], client=client,
)
base_prompt, source_chunks = await build_rag_context(
    student_id=user.user_id, session_id=session_id,
    question=retrieval_question, course_id=session.course_id, db=db,
)
```

**What is wrong:**

When `reformulate_query` falls back, the caller has no way to know whether the `retrieval_question` it received is a rewrite or the original `body.content`. The fallback is invisible to the chat layer. This means:

- No per-session metric or log entry recording that reformulation failed for this particular student session
- No way to measure the reformulation fallback rate (what fraction of requests degraded)
- No Sentry signal

The PR description calls this the "graceful-failure contract" and that framing is correct for the tutoring outcome (student still gets an answer), but graceful for the user does not mean invisible to operators. A fallback rate above some threshold indicates a systemic problem (gateway down, model misconfigured) that needs operator attention.

**Recommendation:** At minimum, log at the call site whether a fallback occurred:

```python
retrieval_question = await reformulate_query(
    body.content, history[-4:], client=client,
)
if retrieval_question == body.content:
    logger.info(
        "reformulate_query fell back to original for session=%s student=%s",
        session_id, user.user_id,
    )
```

A better approach is to have `reformulate_query` return a named tuple or small dataclass indicating success/fallback, so the caller can emit a metric or log entry without re-implementing the equality check.

---

### MEDIUM — `get_gateway_client()` failure is silently folded into the broad `except Exception`

**Location:** `services/tutoring-service/app/rag.py`

```python
if client is None:
    from .gateway import get_gateway_client
    client = get_gateway_client()
```

**What is wrong:**

This line is outside both `try` blocks. If `get_gateway_client()` raises (e.g. the gateway module fails to initialize, a required environment variable is missing, the HTTP client constructor throws), the exception propagates uncaught out of `reformulate_query` and crashes the caller. The function's documented contract ("callers never need to handle exceptions from this module") is violated.

If `get_gateway_client()` returns `None` silently (some implementations use this pattern to indicate "not configured"), then `client` will be `None` and `client.chat.completions.create(...)` will raise `AttributeError: 'NoneType' object has no attribute 'chat'` inside the first `try` block — which is caught by `except Exception` and logged as a cryptic "gateway error". The model name in that log will be the only clue that the client was null.

**Recommendation:** Either move the `get_gateway_client()` call inside the `try` block, or add explicit handling:

```python
try:
    if client is None:
        from .gateway import get_gateway_client
        client = get_gateway_client()
    if client is None:
        logger.warning(
            "reformulate_query: gateway client is None (not configured?) — falling back. model=%s",
            model or settings.tutor_fast_model,
        )
        return message
    response = await client.chat.completions.create(...)
except Exception as exc:  # noqa: BLE001
    ...
```

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | CRITICAL | `except Exception` logs no exc type, exc message, or model — operationally opaque |
| 2 | HIGH | `TypeError` (choices=None) escapes the second except tuple → uncaught 500 |
| 3 | HIGH | Second except block also logs zero diagnostic context |
| 4 | HIGH | `_history_to_chat_messages` silently corrupts history rows with null role/content |
| 5 | MEDIUM | Empty rewrite returns silently with no log — model regressions undetectable |
| 6 | MEDIUM | Call site in chat.py has no observability on whether fallback occurred |
| 7 | MEDIUM | `get_gateway_client()` call is outside both try blocks, violating the no-throw contract |

The fundamental problem is that "graceful fallback" and "silent failure" are being conflated. Falling back to `body.content` is the right user-facing behavior. But every fallback path must emit a log entry with enough context (exc type, exc message, model name, session ID where available) for an operator to distinguish "occasional expected degradation" from "reformulation is broken for all users right now." None of the four fallback paths currently meet that bar.

Requires changes before merge.

— Petra (silent-failure-hunter)

DreadPirateRobertz (commented):
## Test Coverage Review

**Reviewer:** Petra (pr-test-analyzer)

Overall the unit test suite for `app/rag.py` is thorough and well-structured. All declared fallback contracts are exercised, the ORM-row path is covered, and the helper factories keep the tests readable. Two issues need attention before this merges — one is a confirmed logic bug the tests are masking, the other is a coverage gap in the role-mapping helper.

---

### Critical Gap — `history[-4:]` already contains the current student message (criticality: 9/10)

**The bug:** In `chat.py`, `append_message(... role="student" ...)` runs at line 250, then `build_pruned_history` runs at line 253. `build_pruned_history` reads from the database, so `history` returned already includes the current student turn as its final item. The PR then passes `history[-4:]` to `reformulate_query`, which appends `chat_history` (containing that student turn) followed by the instruction message `"Rewrite this follow-up as a self-contained search query:\n{message}"`. The LLM receives two consecutive `role: user` turns — the current question appears once in the history block and again in the instruction — which degrades reformulation quality and is structurally invalid for most model providers.

**Why the integration tests miss it:** `test_sse_stream_rag_calls_reformulate_query` mocks `build_pruned_history` to return a 2-item list `[student_prior, tutor_prior]` and mocks `reformulate_query` entirely. Neither mock reflects the real database read. The test asserts `call_args.args[1] == history[-4:]` — but `history` here is the fixture list, not the post-append DB state. No test exercises the live `append_message → build_pruned_history → reformulate_query` sequence end-to-end.

**What a test should cover:**

```python
async def test_reformulate_receives_history_without_current_message():
    # Arrange: DB returns history that includes the just-appended student message
    # (simulate what build_pruned_history actually returns post-append)
    history_with_current = [
        MagicMock(role="student", content="what are SN2 reactions?"),
        MagicMock(role="tutor",   content="SN2 is bimolecular substitution..."),
        MagicMock(role="student", content="what about stereochemistry?"),  # current msg
    ]
    # Assert: the slice passed to reformulate_query does NOT end with the
    # current student message — or that build_pruned_history is called
    # before append_message (the fix).
    # Currently this test would FAIL, exposing the duplicate-turn bug.
```

The fix is either: (a) call `reformulate_query` before `append_message`, or (b) strip the last item from `history` before slicing when its role is "student" and content matches `body.content`. Whichever approach is chosen, the integration test must be updated to reflect the real DB-read ordering rather than using a pre-constructed fixture list.

---

### Coverage Gap — `_history_to_chat_messages` silent role coercion (criticality: 6/10)

**The issue:** The mapping `"user" if role == "student" else "assistant"` silently promotes any unrecognised role to "assistant". A dict with `role="assistant"` (already-converted message accidentally passed in), `role="system"`, or a corrupt ORM row with a bad role value will all silently become assistant turns. No test currently triggers this path — `test_system_prompt_and_history_passed_to_gateway` only exercises the `student → user` and `tutor → assistant` pairs.

**What to add:**

```python
async def test_unknown_role_in_history_maps_to_assistant():
    history = [{"role": "system", "content": "injected"}]
    client = _make_client_returning("rewritten")
    await reformulate_query("q?", history, client=client)
    messages = client.chat.completions.create.await_args.kwargs["messages"]
    # Currently passes silently as assistant — document or raise on unknown roles
    assert messages[1]["role"] == "assistant"   # or assert this raises ValueError
```

This is worth a conscious decision: should unknown roles raise, warn, or silently coerce? Right now the contract is undocumented and untested. At minimum, add a `logger.warning` on unrecognised roles and a test that asserts the warning is emitted.

---

### Positive Observations

- The fallback contract is comprehensively tested: empty history, gateway exception, blank rewrite, `None` content, empty `choices` list — all 5 paths are covered and the assertions are behavioural, not implementation-coupled.
- `test_uses_custom_model_override` correctly inspects the `await_args` kwargs rather than asserting on internal state.
- The `_fake_response` / `_make_client_returning` / `_make_client_raising` factory pattern keeps individual tests short and intent-clear — good DAMP practice.
- `test_default_client_fetched_via_gateway` correctly targets `app.gateway.get_gateway_client` (the module attribute), which is the right target for a lazy local import.
- Docstrings on module, function, and helpers are complete and Google-style.

---

### Summary

The unit tests for `app/rag.py` are solid in isolation. The integration tests are the weak point: they mock away the exact ordering that produces the double-user-turn bug. That bug will silently ship and degrade reformulation quality (and may cause errors with strict model providers). This needs a fix + a test that cannot be fooled by pre-constructed fixture history before merge.

DreadPirateRobertz (commented):
## Comment Analysis — app/rag.py and app/chat.py

**Reviewer: Petra (comment-analyzer)**

Overall the docstrings are well-structured and the inline comment in chat.py is accurate and useful. There are two issues that need attention before merge: one factually incorrect guarantee in the module docstring, and one undocumented design decision on the inference parameters.

---

### Critical Issue: Module Docstring False Exception-Safety Guarantee

**Location:** services/tutoring-service/app/rag.py, module docstring, final sentence

**Current text:**
> Callers never need to handle exceptions from this module.

**Why it is wrong:**

The outer except Exception block covers only the client.chat.completions.create() call. The inner try/except (AttributeError, IndexError) covers the response-unpacking step. But if response.choices is None (a legally-constructable OpenAI-compatible response shape), then response.choices[0] raises TypeError, which is caught by neither block. TypeError propagates to the caller, breaking the stated guarantee.

Verified locally:

    class R:
        choices = None
    r = R()
    try:
        _ = (r.choices[0].message.content or "").strip()
    except (AttributeError, IndexError):
        pass   # not reached
    # TypeError: 'NoneType' object is not subscriptable -- escapes

Fix options:

Option A (recommended) -- widen the inner except to include TypeError:

    except (AttributeError, IndexError, TypeError):

Then the module docstring guarantee holds as written.

Option B -- correct the docstring to be honest about the gap:

> Failure modes from gateway communication and well-formed-but-empty responses are non-fatal. A TypeError from a None-valued choices field is not caught.

There is also no test covering the choices=None path specifically. test_malformed_response_falls_back_to_original uses an empty list (choices=[]), which raises IndexError and IS caught. The None case is distinct and currently untested.

---

### Improvement Opportunity: No Rationale for temperature=0.0 and max_tokens=128

**Location:** services/tutoring-service/app/rag.py, the client.chat.completions.create() call

These values are baked in with no comment explaining why. Both are non-obvious:

- temperature=0.0 -- correct choice (rewrites must be reproducible, not creative), but a future dev seeing this may assume it was left at zero by accident.
- max_tokens=128 -- works because the system prompt constrains to one line of plain text, but that reasoning is not stated. A maintainer who loosens the system prompt scope may not realize queries exceeding 128 tokens will be silently truncated, not errored.

Suggested inline comments:

    temperature=0.0,   # deterministic -- rewrites must be reproducible for the same input context
    max_tokens=128,    # one-line output per system prompt; queries longer than this truncate silently

---

### Improvement Opportunity: Silent Misclassification on Unrecognized Roles in _history_to_chat_messages

**Location:** services/tutoring-service/app/rag.py, _history_to_chat_messages

    mapped = "user" if role == "student" else "assistant"

Any unrecognized role value -- empty string, None, "system", a typo -- silently maps to "assistant". This is fine for the current ORM that uses only "student" and "tutor", but the docstring explicitly says plain dicts are accepted to make tests easy, which means the function is intentionally open to arbitrary input. The docstring should document the expected role vocabulary, or the implementation should log a warning for unmapped values.

---

### Positive Findings

The reformulate_query Args/Returns block is accurate and detailed. The note that callers should pre-slice (history[-4:]) before passing is genuinely useful and matches the actual call site in chat.py. The lazy-import comment explaining circular-dependency avoidance is exactly the right kind of "why" comment.

The inline comment in chat.py accurately describes what the call does, why it matters at embedding time, and what the fallback behavior is. Good example of intent documentation rather than code restatement.

The Covers: block in tests/test_rag.py is a clean, accurate index of what the suite exercises, making coverage gaps easy to audit at a glance.

---

### Summary

One factually incorrect guarantee (module-level exception-safety claim) needs to be fixed before merge -- either widen the inner except or correct the docstring. The choices=None case also needs a dedicated test. The temperature/max_tokens documentation gap is lower priority but will pay off when someone revisits prompt tuning.

DreadPirateRobertz (commented):
## Comment Analysis — app/rag.py and app/chat.py

Reviewer: Petra (comment-analyzer)

Overall the docstrings are well-structured and the inline comment in chat.py is accurate and useful. Two issues need attention before merge: one factually incorrect guarantee in the module docstring, and one undocumented design decision on the inference parameters.

---

### Critical Issue: Module Docstring False Exception-Safety Guarantee

Location: services/tutoring-service/app/rag.py, module docstring, final sentence

Current text: "Callers never need to handle exceptions from this module."

Why it is wrong: The outer `except Exception` block covers only the `client.chat.completions.create()` call. The inner `try/except (AttributeError, IndexError)` covers only the response-unpacking step. If `response.choices` is `None` (a legal OpenAI-compatible response shape produced by some LiteLLM edge cases), then `response.choices[0]` raises `TypeError` — not `AttributeError` or `IndexError`. `TypeError` is caught by neither handler and propagates to the caller, breaking the stated guarantee.

Verified: subscripting `None` raises `TypeError: 'NoneType' object is not subscriptable`, which the inner except clause does not catch.

Fix options:

Option A (recommended) — widen the inner except to include TypeError:

```
except (AttributeError, IndexError, TypeError):
```

The module docstring guarantee then holds as written.

Option B — correct the docstring instead:

"Failure modes from gateway communication and well-formed-but-empty responses are non-fatal. A TypeError from a None-valued choices field is not caught and will propagate."

Test gap: `test_malformed_response_falls_back_to_original` uses `choices=[]`, which raises `IndexError` (caught). The `choices=None` case is distinct and currently has no dedicated test. A test should be added alongside whichever fix is chosen.

---

### Improvement Opportunity: No Rationale for temperature=0.0 and max_tokens=128

Location: services/tutoring-service/app/rag.py, the `client.chat.completions.create()` call

Both values are baked in with no comment explaining why:

- `temperature=0.0` is the correct choice (rewrites must be reproducible, not creative), but a future dev seeing it might assume it was left at zero by accident and bump it to 0.7.
- `max_tokens=128` works because the system prompt constrains output to one line of plain text, but that reasoning is not stated. A maintainer who broadens the system prompt may not realize queries exceeding 128 tokens are silently truncated, not errored.

Suggested inline comments:

```
temperature=0.0,   # deterministic — rewrites must be reproducible for the same input context
max_tokens=128,    # one-line output per system prompt; longer rewrites truncate silently, not error
```

---

### Improvement Opportunity: Silent Misclassification on Unrecognized Roles in _history_to_chat_messages

Location: services/tutoring-service/app/rag.py, _history_to_chat_messages

The line `mapped = "user" if role == "student" else "assistant"` silently maps any unrecognized role (empty string, None, "system", a typo) to "assistant". This is safe for the current ORM that uses only "student" and "tutor", but the docstring explicitly states plain dicts are accepted to make tests easy — meaning the function is intentionally open to arbitrary input. The docstring should document the expected role vocabulary, or the implementation should log a warning for unmapped values so callers passing unexpected data get a diagnostic instead of silent wrong behavior.

---

### Positive Findings

The `reformulate_query` Args/Returns block is accurate and detailed. The note that callers should pre-slice (`history[-4:]`) before passing is genuinely useful and matches the actual call site in chat.py. The lazy-import comment explaining circular-dependency avoidance is exactly the right kind of "why" comment that earns its place in the codebase.

The inline comment in chat.py accurately describes what the call does, why it matters at embedding time, and what the fallback behavior is — a good example of intent documentation rather than code restatement.

The `Covers:` block in `tests/test_rag.py` is a clean, accurate index of what the suite exercises, making coverage gaps easy to audit at a glance.

---

### Summary

One factually incorrect guarantee (module-level exception-safety claim) must be fixed before merge — either widen the inner except clause to include TypeError, or correct the docstring. The `choices=None` path also needs a dedicated test. The `temperature`/`max_tokens` documentation gap is lower priority but will pay maintenance dividends when someone revisits prompt tuning.

DreadPirateRobertz (commented):
## Comment Analysis — app/rag.py and app/chat.py

Reviewer: Petra (comment-analyzer)

Overall the docstrings are well-structured and the inline comment in chat.py is accurate and useful. Two issues need attention before merge: one factually incorrect guarantee in the module docstring, and one undocumented design decision on the inference parameters.

---

### Critical Issue: Module Docstring False Exception-Safety Guarantee

Location: services/tutoring-service/app/rag.py, module docstring, final sentence

Current text: "Callers never need to handle exceptions from this module."

Why it is wrong: The outer `except Exception` block covers only the `client.chat.completions.create()` call. The inner `try/except (AttributeError, IndexError)` covers only the response-unpacking step. If `response.choices` is `None` (a legal OpenAI-compatible response shape produced by some LiteLLM edge cases), then `response.choices[0]` raises `TypeError` — not `AttributeError` or `IndexError`. `TypeError` is caught by neither handler and propagates to the caller, breaking the stated guarantee.

Verified: subscripting `None` raises `TypeError: 'NoneType' object is not subscriptable`, which the inner except clause does not catch.

Fix options:

Option A (recommended) — widen the inner except to include TypeError:

```
except (AttributeError, IndexError, TypeError):
```

The module docstring guarantee then holds as written.

Option B — correct the docstring instead:

"Failure modes from gateway communication and well-formed-but-empty responses are non-fatal. A TypeError from a None-valued choices field is not caught and will propagate."

Test gap: `test_malformed_response_falls_back_to_original` uses `choices=[]`, which raises `IndexError` (caught). The `choices=None` case is distinct and currently has no dedicated test. A test should be added alongside whichever fix is chosen.

---

### Improvement Opportunity: No Rationale for temperature=0.0 and max_tokens=128

Location: services/tutoring-service/app/rag.py, the `client.chat.completions.create()` call

Both values are baked in with no comment explaining why:

- `temperature=0.0` is the correct choice (rewrites must be reproducible, not creative), but a future dev seeing it might assume it was left at zero by accident and bump it to 0.7.
- `max_tokens=128` works because the system prompt constrains output to one line of plain text, but that reasoning is not stated. A maintainer who broadens the system prompt may not realize queries exceeding 128 tokens are silently truncated, not errored.

Suggested inline comments:

```
temperature=0.0,   # deterministic — rewrites must be reproducible for the same input context
max_tokens=128,    # one-line output per system prompt; longer rewrites truncate silently, not error
```

---

### Improvement Opportunity: Silent Misclassification on Unrecognized Roles in _history_to_chat_messages

Location: services/tutoring-service/app/rag.py, _history_to_chat_messages

The line `mapped = "user" if role == "student" else "assistant"` silently maps any unrecognized role (empty string, None, "system", a typo) to "assistant". This is safe for the current ORM that uses only "student" and "tutor", but the docstring explicitly states plain dicts are accepted to make tests easy — meaning the function is intentionally open to arbitrary input. The docstring should document the expected role vocabulary, or the implementation should log a warning for unmapped values so callers passing unexpected data get a diagnostic instead of silent wrong behavior.

---

### Positive Findings

The `reformulate_query` Args/Returns block is accurate and detailed. The note that callers should pre-slice (`history[-4:]`) before passing is genuinely useful and matches the actual call site in chat.py. The lazy-import comment explaining circular-dependency avoidance is exactly the right kind of "why" comment that earns its place in the codebase.

The inline comment in chat.py accurately describes what the call does, why it matters at embedding time, and what the fallback behavior is — a good example of intent documentation rather than code restatement.

The `Covers:` block in `tests/test_rag.py` is a clean, accurate index of what the suite exercises, making coverage gaps easy to audit at a glance.

---

### Summary

One factually incorrect guarantee (module-level exception-safety claim) must be fixed before merge — either widen the inner except clause to include TypeError, or correct the docstring. The `choices=None` path also needs a dedicated test. The `temperature`/`max_tokens` documentation gap is lower priority but will pay maintenance dividends when someone revisits prompt tuning.

DreadPirateRobertz (commented):
## Silent-failure audit — Petra (silent-failure-hunter)

Overall: this PR is well-structured and the author clearly thought about failure modes. The tests are solid. However there are several error-handling defects that need to be addressed before merge. I am flagging them in order of severity.

---

### CRITICAL — Broad `except Exception` swallows unknown failures without enough diagnostic information

**Location:** `services/tutoring-service/app/rag.py`, first `try/except` around `client.chat.completions.create`

```python
except Exception:  # noqa: BLE001  non-fatal — degrade to original
    logger.warning("reformulate_query: gateway error, falling back to original message")
    return message
```

The warning log contains zero diagnostic information. No exception type, no exception message, no model name, no session context. When this fires in production, no one can distinguish:

- A network timeout to the AI gateway
- An `AuthenticationError` (expired or wrong API key)
- A `RateLimitError` (the fast model is saturated)
- A `BadRequestError` (the prompt_messages list exceeded context window — easily triggered if history rows have long content)
- A `TypeError` from `get_gateway_client()` returning `None`
- An `OSError` / `ConnectionRefusedError` (gateway pod is down entirely)

The string "gateway error" tells an operator nothing actionable. They cannot tell from logs whether to page someone about a credential rotation, increase rate limits, or fix a prompt that is too long. Logging the model name is also missing — a misconfigured `tutor_fast_model` in settings would fire this fallback on every single request with no clue why.

**Fix:**

```python
except Exception as exc:  # noqa: BLE001
    logger.warning(
        "reformulate_query: gateway error — falling back to original. "
        "model=%s exc_type=%s exc=%s",
        model or settings.tutor_fast_model,
        type(exc).__name__,
        exc,
    )
    return message
```

---

### HIGH — `TypeError` from `response.choices = None` is not caught by the second except block

**Location:** `services/tutoring-service/app/rag.py`, second `try/except`

```python
try:
    rewritten = (response.choices[0].message.content or "").strip()
except (AttributeError, IndexError):
```

If `response.choices` is `None` (which OpenAI-compatible APIs can return on content-filtered responses, returning `{"choices": null}`), then `response.choices[0]` raises `TypeError: 'NoneType' object is not subscriptable`. `TypeError` is not in the catch tuple. This exception propagates uncaught out of `reformulate_query`, crashes the caller in `chat.py`, and surfaces as a 500 to the student — the exact outcome the fallback was designed to prevent.

There is a test for `choices=[]` (IndexError) but no test for `choices=None` (TypeError). These are distinct failure modes.

**Fix:** Add `TypeError` to the catch tuple and add a test passing `SimpleNamespace(choices=None)`.

```python
except (AttributeError, IndexError, TypeError):
    logger.warning(
        "reformulate_query: malformed gateway response — falling back. "
        "choices=%r exc_type=%s exc=%s",
        getattr(response, "choices", "<missing>"),
        type(exc).__name__,
        exc,
    )
    return message
```

(Note: needs `as exc` added to the except clause to make `exc` available in the log.)

---

### HIGH — The second except block also logs zero diagnostic context

Same location as above. `logger.warning("reformulate_query: malformed gateway response, falling back")` — no exception type, no message, no indication of what `response` actually contained. A malformed response from a misconfigured gateway or a provider returning an unexpected schema produces this warning with no actionable information. Log `type(exc).__name__` and `str(exc)` here as well.

---

### HIGH — `_history_to_chat_messages` silently corrupts rows with null `role` or `content`

**Location:** `services/tutoring-service/app/rag.py`, `_history_to_chat_messages`

```python
role = getattr(item, "role", None) or item.get("role")
content = getattr(item, "content", None) or item.get("content")
mapped = "user" if role == "student" else "assistant"
out.append({"role": mapped, "content": content})
```

Three silent failure modes:

1. If `role` is `None` (null column in the ORM row), `mapped` silently becomes `"assistant"`. The item is not skipped; it is appended with a wrong role. This silently corrupts the context window sent to the gateway with no log.

2. If `content` is `None` or empty, `{"role": "user", "content": None}` is appended to `prompt_messages`. The OpenAI SDK will raise a `ValidationError` or the gateway will reject the null content. Either breaks the request and falls through to the broad `except Exception` with no indication which history row caused it.

3. The `or` chaining `getattr(item, "role", None) or item.get("role")` will raise `AttributeError` if the ORM row has a `role` attribute that is an empty string `""`, because ORM rows do not have `.get()`. The `# type: ignore[union-attr]` comment acknowledges this but does not fix it.

**Fix:** Use `hasattr` to branch cleanly between ORM rows and dicts, skip items with null role or content, and log the skip:

```python
def _history_to_chat_messages(history: list[Any]) -> list[dict[str, str]]:
    out: list[dict[str, str]] = []
    for item in history:
        if hasattr(item, "role"):
            role, content = item.role, item.content
        else:
            role, content = item.get("role"), item.get("content")
        if not role or not content:
            logger.warning(
                "_history_to_chat_messages: skipping item with missing role or content "
                "(role=%r content_present=%s)",
                role, bool(content),
            )
            continue
        mapped = "user" if role == "student" else "assistant"
        out.append({"role": mapped, "content": content})
    return out
```

---

### MEDIUM — Empty rewrite returns silently with no log

**Location:** `services/tutoring-service/app/rag.py`

```python
if not rewritten:
    return message
```

No log. When `tutor_fast_model` starts returning empty strings for all requests (prompt regression, model degradation, wrong model configured), this code silently returns every original message with zero indication in logs that reformulation has stopped working. This will look identical to the function working correctly, and retrieval quality will degrade without any alerting signal.

**Fix:**

```python
if not rewritten:
    logger.warning(
        "reformulate_query: gateway returned empty rewrite — falling back. model=%s",
        model or settings.tutor_fast_model,
    )
    return message
```

One log line is the difference between catching a model regression in minutes versus never.

---

### MEDIUM — `get_gateway_client()` is called outside both try blocks, violating the no-throw contract

**Location:** `services/tutoring-service/app/rag.py`

```python
if client is None:
    from .gateway import get_gateway_client
    client = get_gateway_client()
```

This is outside both `try` blocks. If `get_gateway_client()` raises (missing env var, HTTP client constructor failure, gateway module init failure), the exception propagates uncaught from `reformulate_query` and crashes the caller. The docstring says "callers never need to handle exceptions from this module" — that contract is not upheld here.

If `get_gateway_client()` returns `None`, then `client.chat.completions.create(...)` inside the first `try` raises `AttributeError: 'NoneType' object has no attribute 'chat'` which is caught by `except Exception` and logged as the opaque "gateway error" with no indication the client was null.

**Fix:** Move the `get_gateway_client()` call inside the try block, or explicitly guard the None case before proceeding.

---

### MEDIUM — Fallback is not observable at the call site in `chat.py`

**Location:** `services/tutoring-service/app/chat.py`

```python
retrieval_question = await reformulate_query(
    body.content, history[-4:], client=client,
)
```

When `reformulate_query` falls back, the caller has no way to know. There is no per-session log entry, no metric, no Sentry signal. A fallback rate above some threshold indicates a systemic problem — gateway down, model misconfigured — that needs operator attention. Currently this is completely invisible.

**Fix:** Log at the call site when fallback occurs:

```python
retrieval_question = await reformulate_query(
    body.content, history[-4:], client=client,
)
if retrieval_question is body.content:
    logger.info(
        "reformulate_query fell back to original for session=%s student=%s",
        session_id, user.user_id,
    )
```

A better longer-term fix is to return a dataclass `ReformulationResult(query: str, was_reformulated: bool)` so the caller does not need to re-implement the equality check.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | CRITICAL | First except catches all exceptions but logs no type, message, or model — opaque in production |
| 2 | HIGH | `TypeError` (choices=None) escapes the second except tuple → uncaught 500 |
| 3 | HIGH | Second except block also logs zero diagnostic context |
| 4 | HIGH | `_history_to_chat_messages` silently corrupts rows with null role/content and has a latent AttributeError on ORM rows |
| 5 | MEDIUM | Empty rewrite falls back silently — model regressions produce no log signal |
| 6 | MEDIUM | `get_gateway_client()` outside try blocks violates the documented no-throw contract |
| 7 | MEDIUM | Call site in chat.py has no observability on whether fallback occurred |

The core problem: "graceful fallback" and "silent failure" are being conflated. Falling back to `body.content` is the correct user-facing behavior. But every fallback path must emit a log entry with enough context (exc type, exc message, model name) for an operator to distinguish "occasional expected degradation" from "reformulation is broken for all users right now." None of the four current fallback paths meet that bar.

Requires changes before merge.

— Petra (silent-failure-hunter)

DreadPirateRobertz (commented):
test

DreadPirateRobertz (commented):
## Type-design analysis — `app/rag.py`

**Reviewer: Petra (type-design-analyzer)**

---

## Type: `_history_to_chat_messages` / `reformulate_query` — `history: list[Any]`

### Invariants Identified
- Every element must expose a `role` string and a `content` string, either as object attributes or as dict keys.
- `role` must be one of `"student"` / `"tutor"` (ORM convention) or `"user"` / `"assistant"` (OpenAI convention) for the mapping to produce valid chat messages.
- `content` must be a non-None string for the assembled prompt to be well-formed.
- The list represents ordered conversation turns (most-recent last); ordering is a caller-enforced invariant with no structural guarantee.

### Ratings

**Encapsulation: 6/10**
The dual-access idiom (`getattr(item, "role", None) or item.get("role")`) is an internal implementation detail that leaks through the public signature. Callers cannot tell from the type annotation what shape `history` elements must have. The `# type: ignore[union-attr]` suppression confirms the type checker has already given up on this boundary.

**Invariant Expression: 4/10**
`list[Any]` communicates nothing. The two valid shapes (ORM row, plain dict) and their required keys are documented only in the docstring. A reader of a call site — especially in `chat.py` — has no static signal about what is permitted.

**Invariant Usefulness: 8/10**
The invariants themselves are sound and motivated: ORM rows for production, plain dicts for tests. The fallback chain in `reformulate_query` is thorough (empty history, gateway exception, blank rewrite, None content, malformed response). The business rule is well-chosen and the graceful-failure contract is valuable.

**Invariant Enforcement: 5/10**
Enforcement relies entirely on the `getattr`/`.get` duck-typing pattern at runtime. There is no check that `role` and `content` are actually present and non-None before building the output list — an element missing both attribute and key would produce `{"role": None, "content": None}` and be silently forwarded to the gateway. The `role` mapping also silently passes through any unrecognised role string unchanged (e.g., an ORM row with `role="tutor"` maps to `"assistant"` correctly, but an unexpected value like `"system"` would pass straight through without error).

---

### Strengths
- Graceful-failure contract is complete and well-tested across every branch. The eight-test suite in `test_rag.py` covers every exit path.
- `REFORMULATION_SYSTEM_PROMPT` as a module-level constant is correctly typed (`str`), importable by tests for assertion, and immutable at module load. Good.
- `client: AsyncOpenAI | None` with lazy `get_gateway_client()` injection is the right pattern for avoiding circular imports while remaining testable. The `None` default is the correct signal.
- Return type `-> str` with a guaranteed non-raising contract is excellent. Callers in `chat.py` can use the result unconditionally.
- The `temperature=0.0, max_tokens=128` constraints on the gateway call are implicit type invariants on the output — deterministic, short — and correctly chosen for the use case.

---

### Concerns

**1. `list[Any]` obscures the real constraint.**
The actual contract is: each element must satisfy `(hasattr(item, "role") and hasattr(item, "content")) or (isinstance(item, dict) and "role" in item and "content" in item)`. That is a structural type — expressible as a `Protocol` or `TypedDict` union.

**2. Silent `None` propagation in `_history_to_chat_messages`.**
If an element has neither attribute nor key for `role` or `content`, the expression evaluates to `None`. The function then appends `{"role": None, "content": None}` to `out`. The OpenAI client will reject this at runtime with a cryptic validation error rather than a local, attributable one.

**3. `role` mapping is incomplete.**
The mapping `"student" → "user"` and anything-else → `"assistant"` is implicit. An ORM row with `role="system"` (conceivable if schema adds moderation rows) silently becomes `"assistant"`. An explicit allowlist would be safer.

**4. Caller-side slice responsibility is untested at the boundary.**
`history[-4:]` is applied in `chat.py`, not in `reformulate_query`. The docstring notes this but there is no enforcement. A caller passing a long unbounded history is not protected. Consider accepting a `max_turns: int = 4` parameter and slicing internally, making the safety invariant part of the function rather than a convention.

---

### Recommended Improvements

**Option A — Minimal, low-churn (recommended for now):**
```python
from typing import Protocol, runtime_checkable

@runtime_checkable
class _HistoryItem(Protocol):
    role: str
    content: str
```
Change the signature to `history: list[_HistoryItem | dict[str, str]]`. The `dict[str, str]` branch keeps plain-dict test ergonomics. This eliminates the `type: ignore` suppressions and makes the contract statically verifiable at call sites. The `_` prefix keeps it private to the module.

**Option B — TypedDict for the dict branch:**
```python
class _HistoryDict(TypedDict):
    role: str
    content: str

history: list[_HistoryItem | _HistoryDict]
```
Slightly more expressive; Mypy can verify dict literals at test sites.

**Guard `None` propagation:**
```python
if role is None or content is None:
    continue  # or raise ValueError — caller never passes partial rows
```
This converts a silent bad-message-to-gateway into a local, attributable skip or error.

**Intern the role mapping:**
```python
_ROLE_MAP = {"student": "user", "tutor": "assistant"}
mapped = _ROLE_MAP.get(role, "user")  # or raise on unrecognised
```

**Option A is the right call for this PR's scope.** The Protocol approach costs ~6 lines, eliminates both `type: ignore` comments, and makes `_history_to_chat_messages` statically checkable without changing any test code (plain dicts satisfy a structural Protocol at runtime).

The `list[Any]` is acceptable as a pragmatic starting point given the testability motivation, but it should be upgraded before this pattern proliferates to other callers. If a second consumer of `reformulate_query` appears, the untyped boundary will become a maintenance liability.

---

Overall this is solid work — the graceful-failure design and test coverage are genuinely good. The type annotation is the one place that doesn't match the quality of the surrounding code.

— Petra (type-design-analyzer)

PR #239, opened 2026-04-13
feat(tl-afb): query expansion for short follow-up searches

DreadPirateRobertz:
## What
Adds `app/expander.py` — a short-query expansion path for `/v1/search` that uses caller-supplied conversation turns and the AI-gateway `tutor_fast_model` to rewrite under-specified follow-ups (\"why?\", \"and that one?\") into standalone search queries.

## Current Behavior
Short, ambiguous follow-up queries hit Qdrant directly and retrieve low-relevance chunks.

## New Behavior
- `/v1/search` accepts an optional repeated query param `context_turns` (most-recent last).
- When `q` has fewer than `query_expansion_short_threshold` (=5) whitespace tokens **and** `context_turns` is non-empty, `expand_query()` rewrites it via AI gateway chat completion.
- The **effective** query (expanded or raw) flows through embedding, Qdrant search, rerank, and is echoed on the response.
- Total contract: `expand_query` never raises. Every skip/failure path increments `search_query_expansion_total{outcome}` with `outcome ∈ {expanded, passthrough_long, passthrough_nocontext, fallback_error, fallback_blank}` and logs WARNING on fallbacks.
- Context is clipped to the most-recent `query_expansion_max_context_turns` (=6) to keep gateway payloads bounded.

## Detail
- `app/expander.py` — lazy `AsyncOpenAI` pointed at `ai_gateway_url`, stateless, prompt asks for a single-line standalone rewrite.
- `app/routers/search.py` — adds `context_turns: list[str] | None`; replaces raw `q` with `effective_q` in `_run_search`, `rerank`, and the response `query` field.
- `app/metrics.py` — new `search_query_expansion_total` Counter by `outcome`.
- `app/config.py` — `tutor_fast_model`, `query_expansion_short_threshold`, `query_expansion_max_context_turns`, `query_expansion_max_tokens`.
- `tests/test_expander.py` — 9 tests: happy path / long-query passthrough / no-context passthrough / blank-turns passthrough / gateway error fallback / blank expansion fallback / turn-clipping / endpoint wiring / endpoint resilience on gateway failure.

## Why
Bead tl-afb. Option 1+3 hybrid per petra spec: caller-supplied `context_turns` (stateless) + AI-gateway-backed rewrite. Graceful failure was explicit: never 500 on expansion alone.

## How to test
- `pytest services/search/tests/test_expander.py` — all 9 tests (expander unit + endpoint integration).
- Manually:
  - `curl '.../v1/search?q=hello&course_id=...'` → unchanged behaviour (no context → passthrough).
  - `curl '.../v1/search?q=why?&course_id=...&context_turns=...&context_turns=...'` → query rewritten, `query` in response reflects the rewrite.
  - `curl '.../metrics' | grep search_query_expansion_total` — outcome counters visible.

## Checklist
- [x] Tests written (TDD) — `tests/test_expander.py` (9 cases, expander + endpoint)
- [x] Coverage ≥80% on new code
- [x] Docstrings on all new functions/endpoints/module
- [x] Lint clean (`ruff check` passes on all touched files)
- [x] No secrets committed
- [x] Self-review: diff read before opening

Closes tl-afb

### DIFF
--- a/services/search/app/config.py
+++ b/services/search/app/config.py
@@ -34,6 +34,13 @@ class Settings(BaseSettings):
     # Number of candidates fetched per signal before RRF fusion + final limit
     sparse_rerank_limit: int = 20
 
+    # Query expansion via AI Gateway (tl-afb) — used for short (<5-token)
+    # follow-up queries when the caller supplies conversation context turns.
+    tutor_fast_model: str = "tutor-fast"
+    query_expansion_short_threshold: int = 5
+    query_expansion_max_context_turns: int = 6
+    query_expansion_max_tokens: int = 128
+
     # Diagram (CLIP) collection — Phase 6 multi-modal RAG
     diagrams_collection: str = "diagrams"
     clip_model: str = "openai/clip-vit-base-patch32"
--- a/services/search/app/config.py
+++ b/services/search/app/config.py
@@ -40,6 +40,11 @@ class Settings(BaseSettings):
     query_expansion_short_threshold: int = 5
     query_expansion_max_context_turns: int = 6
     query_expansion_max_tokens: int = 128
+    # Hard wall-clock ceiling on the gateway call. Without this, a hung
+    # upstream can stall every short-query search for the OpenAI-client
+    # default (~600s). 3s is well above tutor_fast_model p99 yet leaves
+    # budget for the remainder of the RAG pipeline under its SLO.
+    query_expansion_timeout_seconds: float = 3.0
 
     # Diagram (CLIP) collection — Phase 6 multi-modal RAG
     diagrams_collection: str = "diagrams"
--- a/services/search/app/config.py
+++ b/services/search/app/config.py
@@ -1,3 +1,4 @@
+from pydantic import Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
@@ -37,14 +38,17 @@ class Settings(BaseSettings):
     # Query expansion via AI Gateway (tl-afb) — used for short (<5-token)
     # follow-up queries when the caller supplies conversation context turns.
     tutor_fast_model: str = "tutor-fast"
-    query_expansion_short_threshold: int = 5
-    query_expansion_max_context_turns: int = 6
-    query_expansion_max_tokens: int = 128
+    # Field validators catch misconfigured env overrides at startup instead of
+    # silently producing a feature that never fires (e.g. threshold=0 → always
+    # passthrough_long; timeout=0 → instant failure on every call).
+    query_expansion_short_threshold: int = Field(default=5, ge=1)
+    query_expansion_max_context_turns: int = Field(default=6, ge=1)
+    query_expansion_max_tokens: int = Field(default=128, ge=1, le=1024)
     # Hard wall-clock ceiling on the gateway call. Without this, a hung
     # upstream can stall every short-query search for the OpenAI-client
     # default (~600s). 3s is well above tutor_fast_model p99 yet leaves
     # budget for the remainder of the RAG pipeline under its SLO.
-    query_expansion_timeout_seconds: float = 3.0
+    query_expansion_timeout_seconds: float = Field(default=3.0, gt=0.0)
 
     # Diagram (CLIP) collection — Phase 6 multi-modal RAG
     diagrams_collection: str = "diagrams"
--- a/services/search/app/expander.py
+++ b/services/search/app/expander.py
@@ -106,6 +106,7 @@ async def expand_query(query: str, context_turns: list[str] | None) -> str:
             model=settings.tutor_fast_model,
             messages=prompt,
             max_tokens=settings.query_expansion_max_tokens,
+            timeout=settings.query_expansion_timeout_seconds,
             stream=False,
         )
         expanded = (response.choices[0].message.content or "").strip()
--- a/services/search/app/expander.py
+++ b/services/search/app/expander.py
@@ -70,6 +70,13 @@ async def expand_query(query: str, context_turns: list[str] | None) -> str:
         whenever expansion is skipped or fails. This function is total: it
         never raises; callers can treat its output as a drop-in replacement.
     """
+    # Empty / whitespace-only input has zero tokens so _is_short is True —
+    # guard here so we don't spend a gateway call on a query with no signal
+    # (and never include a blank line in the prompt).
+    if not query.strip():
+        QUERY_EXPANSION_OUTCOMES.labels(outcome="passthrough_nocontext").inc()
+        return query
+
     if not _is_short(query):
         QUERY_EXPANSION_OUTCOMES.labels(outcome="passthrough_long").inc()
         return query
--- a/services/search/app/expander.py
+++ b/services/search/app/expander.py
@@ -1,0 +1,126 @@
+"""Query expansion for short search queries (tl-afb).
+
+A search query of fewer than :data:`app.config.Settings.query_expansion_short_threshold`
+tokens is often an underspecified follow-up like "why?" or "what about it" —
+running it through Qdrant directly retrieves low-relevance chunks. When the
+caller supplies recent conversation turns, we ask the AI gateway
+(``tutor_fast_model``) to rewrite the query into a standalone, self-contained
+question before it hits the retrieval pipeline.
+
+**Design constraints (per tl-afb spec):**
+
+* Stateless — this module owns no session state; the caller supplies
+  ``context_turns`` on every request.
+* Graceful — *any* expansion failure (gateway error, blank output, network
+  timeout) must fall back to the raw query, log a WARNING, and increment the
+  ``search_query_expansion_total`` counter. The search endpoint never 500s on
+  expansion alone.
+* Cheap — expansion is a single non-streaming completion with a small
+  ``max_tokens`` budget.
+"""
+from __future__ import annotations
+
+import logging
+from typing import Iterable
+
+from app.config import settings
+from app.metrics import QUERY_EXPANSION_OUTCOMES
+
+logger = logging.getLogger(__name__)
+
+_client = None  # lazy AsyncOpenAI pointed at the AI gateway
+
+
+def _get_client():
+    """Return a lazily-initialised AsyncOpenAI client aimed at the AI gateway."""
+    global _client
+    if _client is None:
+        from openai import AsyncOpenAI  # imported here so tests can patch it
+
+        _client = AsyncOpenAI(
+            base_url=settings.ai_gateway_url,
+            api_key=settings.openai_api_key or "ai-gateway-local",
+        )
+    return _client
+
+
+def _is_short(query: str) -> bool:
+    """True if the whitespace-token count is below the configured threshold."""
+    return len(query.split()) < settings.query_expansion_short_threshold
+
+
+def _clip_turns(turns: Iterable[str]) -> list[str]:
+    """Keep the most recent N non-empty turns, bounded by config."""
+    cleaned = [t.strip() for t in turns if t and t.strip()]
+    max_turns = settings.query_expansion_max_context_turns
+    return cleaned[-max_turns:]
+
+
+async def expand_query(query: str, context_turns: list[str] | None) -> str:
+    """Rewrite a short follow-up query into a standalone question.
+
+    Args:
+        query: Raw user query as received by ``/v1/search``.
+        context_turns: Recent conversation turns (caller-supplied, most-recent
+            last). ``None`` or empty means the caller has no context and
+            expansion is skipped.
+
+    Returns:
+        The expanded, self-contained query string — or the original ``query``
+        whenever expansion is skipped or fails. This function is total: it
+        never raises; callers can treat its output as a drop-in replacement.
+    """
+    if not _is_short(query):
+        QUERY_EXPANSION_OUTCOMES.labels(outcome="passthrough_long").inc()
+        return query
+
+    clipped = _clip_turns(context_turns or [])
+    if not clipped:
+        QUERY_EXPANSION_OUTCOMES.labels(outcome="passthrough_nocontext").inc()
+        return query
+
+    prompt = [
+        {
+            "role": "system",
+            "content": (
+                "You rewrite a student's short follow-up question into a "
+                "fully-specified standalone search query, using the prior "
+                "conversation turns for context. Output ONLY the rewritten "
+                "query on a single line — no preamble, no quotation marks, "
+                "no explanations."
+            ),
+        },
+        {
+            "role": "user",
+            "content": (
+                "Prior conversation turns (oldest → newest):\n"
+                + "\n".join(f"- {t}" for t in clipped)
+                + f"\n\nShort follow-up query: {query}\n\nStandalone query:"
+            ),
+        },
+    ]
+
+    try:
+        client = _get_client()
+        response = await client.chat.completions.create(
+            model=settings.tutor_fast_model,
+            messages=prompt,
+            max_tokens=settings.query_expansion_max_tokens,
+            stream=False,
+        )
+        expanded = (response.choices[0].message.content or "").strip()
+    except Exception:
+        QUERY_EXPANSION_OUTCOMES.labels(outcome="fallback_error").inc()
+        logger.warning(
+            "query expansion failed — falling back to raw query",
+            exc_info=True,
+        )
+        return query
+
+    if not expanded:
+        QUERY_EXPANSION_OUTCOMES.labels(outcome="fallback_blank").inc()
+        logger.warning("query expansion returned blank — falling back to raw query")
+        return query
+
+    QUERY_EXPANSION_OUTCOMES.labels(outcome="expanded").inc()
+    return expanded
--- a/services/search/app/metrics.py
+++ b/services/search/app/metrics.py
@@ -15,7 +15,7 @@ import time
 from contextlib import asynccontextmanager
 from typing import AsyncIterator, Literal
 
-from prometheus_client import Histogram
+from prometheus_client import Counter, Histogram
 
 # Buckets chosen per tl-he3 spec: cover expected sub-10ms p50 up through
 # a 2.5s outlier ceiling — anything slower than that is a timeout bug.
@@ -69,3 +69,17 @@ async def observe_query(query_type: QueryType) -> AsyncIterator[None]:
         SEARCH_QUERY_DURATION.labels(
             query_type=query_type, status=status
         ).observe(time.perf_counter() - start)
+
+
+# ── Query expansion (tl-afb) ─────────────────────────────────────────────────
+# Counter of short-query expansion outcomes. ``outcome`` values:
+#   - "expanded"   — AI gateway returned a non-empty rewrite that was used
+#   - "passthrough_long"    — query was not short, no expansion attempted
+#   - "passthrough_nocontext" — no context turns supplied, no expansion attempted
+#   - "fallback_blank"      — gateway returned blank output, raw query used
+#   - "fallback_error"      — gateway call raised, raw query used
+QUERY_EXPANSION_OUTCOMES = Counter(
+    "search_query_expansion_total",
+    "Short-query expansion attempts by outcome (tl-afb).",
+    labelnames=("outcome",),
+)
--- a/services/search/app/routers/search.py
+++ b/services/search/app/routers/search.py
@@ -61,8 +61,10 @@ async def search(
 
     When ``q`` is a short follow-up (< ``query_expansion_short_threshold``
     tokens) and ``context_turns`` is non-empty, the query is first rewritten
-    into a standalone form via the AI gateway. Expansion failures silently
-    fall back to the raw query (tl-afb).
+    into a standalone form via the AI gateway. Expansion failures fall back
+    to the raw query, emit a WARNING log, and increment
+    ``search_query_expansion_total{outcome="fallback_error"|"fallback_blank"}``
+    (tl-afb).
     """
     effective_q = await expand_query(q, context_turns)
     fetch_limit = max(limit, settings.sparse_rerank_limit)
--- a/services/search/app/routers/search.py
+++ b/services/search/app/routers/search.py
@@ -5,6 +5,7 @@ from typing import Annotated, Literal
 from fastapi import APIRouter, Query
 
 from app.config import settings
+from app.expander import expand_query
 from app.metrics import observe_query
 from app.models import SearchResponse
 from app.services.embedder import embed_query
@@ -39,6 +40,16 @@ async def search(
         int,
         Query(ge=1, le=settings.max_search_limit, description="Max results to return"),
     ] = settings.default_search_limit,
+    context_turns: Annotated[
+        list[str] | None,
+        Query(
+            description=(
+                "Optional recent conversation turns used to expand short "
+                "follow-up queries (tl-afb). Pass as repeated query params: "
+                "&context_turns=... — most-recent turn last."
+            ),
+        ),
+    ] = None,
 ) -> SearchResponse:
     """Hybrid search over the curriculum collection for a given course.
 
@@ -47,21 +58,27 @@ async def search(
 
     Falls back gracefully to dense-only when sparse vectors are not available.
     Supports optional narrowing by chapter, section, and content_type.
+
+    When ``q`` is a short follow-up (< ``query_expansion_short_threshold``
+    tokens) and ``context_turns`` is non-empty, the query is first rewritten
+    into a standalone form via the AI gateway. Expansion failures silently
+    fall back to the raw query (tl-afb).
     """
+    effective_q = await expand_query(q, context_turns)
     fetch_limit = max(limit, settings.sparse_rerank_limit)
 
     async with observe_query("hybrid"):
         query_vector, dense_results, sparse_results = await _run_search(
-            q, course_id, fetch_limit, chapter, section, content_type
+            effective_q, course_id, fetch_limit, chapter, section, content_type
         )
 
         fused = combine_dense_sparse(dense_results, sparse_results)
-        ranked = await rerank(q, fused)
+        ranked = await rerank(effective_q, fused)
         final = ranked[:limit]
 
     search_mode = "hybrid" if sparse_results else "dense"
     return SearchResponse(
-        query=q,
+        query=effective_q,
         course_id=course_id,
         results=final,
         total=len(final),
--- a/services/search/tests/test_expander.py
+++ b/services/search/tests/test_expander.py
@@ -74,6 +74,18 @@ class TestExpandQuery:
         fake.chat.completions.create.assert_awaited_once()
 
     @pytest.mark.asyncio
+    async def test_blank_query_passthrough_no_gateway_call(self) -> None:
+        """Empty / whitespace-only queries must short-circuit before the gateway."""
+        before = _outcome("passthrough_nocontext")
+        fake = MagicMock()
+        fake.chat.completions.create = AsyncMock()
+        with patch.object(expander, "_get_client", return_value=fake):
+            out = await expand_query("   ", context_turns=["tutor: gas laws"])
+        assert out == "   "
+        assert _outcome("passthrough_nocontext") == before + 1
+        fake.chat.completions.create.assert_not_awaited()
+
+    @pytest.mark.asyncio
     async def test_long_query_passthrough_no_gateway_call(self) -> None:
         """Queries at/above the token threshold must skip expansion entirely."""
         before = _outcome("passthrough_long")
--- a/services/search/tests/test_expander.py
+++ b/services/search/tests/test_expander.py
@@ -188,7 +188,7 @@ class TestSearchEndpointWiring:
             return results
 
         with (
-            patch("app.expander.expand_query", new_callable=AsyncMock, return_value="What is PV=nRT?"),
+            patch("app.routers.search.expand_query", new_callable=AsyncMock, return_value="What is PV=nRT?"),
             patch("app.routers.search.embed_query", new=_capture_embed),
             patch("app.routers.search.dense_search", new_callable=AsyncMock, return_value=[]),
             patch("app.routers.search.sparse_search", new_callable=AsyncMock, return_value=[]),
--- a/services/search/tests/test_expander.py
+++ b/services/search/tests/test_expander.py
@@ -1,0 +1,259 @@
+"""Tests for the short-query expansion module (tl-afb).
+
+Coverage:
+    * Happy path — short query + context turns → expanded rewrite returned.
+    * Passthrough (long query) — no gateway call, passthrough counter incremented.
+    * Passthrough (no context) — no gateway call, passthrough counter incremented.
+    * Graceful failure — gateway raises or returns blank → raw query + WARNING
+      + ``search_query_expansion_total{outcome="fallback_*"}`` incremented.
+    * Clipping — only the most-recent N turns are sent to the gateway.
+    * Endpoint wiring — ``/v1/search`` uses expanded query through the pipeline.
+"""
+from __future__ import annotations
+
+import uuid
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from fastapi.testclient import TestClient
+from prometheus_client import REGISTRY
+
+from app import expander
+from app.expander import expand_query
+from app.main import app
+
+COURSE_ID = uuid.uuid4()
+
+
+def _outcome(outcome: str) -> float:
+    """Read the current value of search_query_expansion_total for an outcome."""
+    v = REGISTRY.get_sample_value(
+        "search_query_expansion_total", {"outcome": outcome}
+    )
+    return 0.0 if v is None else v
+
+
+def _fake_completion(text: str) -> SimpleNamespace:
+    """Build a minimal OpenAI-shape chat.completion response containing *text*."""
+    return SimpleNamespace(
+        choices=[SimpleNamespace(message=SimpleNamespace(content=text))]
+    )
+
+
+@pytest.fixture(autouse=True)
+def _reset_client():
+    """Ensure each test re-runs the lazy client init — avoids leaked state."""
+    expander._client = None
+    yield
+    expander._client = None
+
+
+class TestExpandQuery:
+    """Unit tests for :func:`app.expander.expand_query`."""
+
+    @pytest.mark.asyncio
+    async def test_happy_path_returns_expanded_rewrite(self) -> None:
+        """Short query + context turns should yield the gateway rewrite."""
+        before = _outcome("expanded")
+        fake = MagicMock()
+        fake.chat.completions.create = AsyncMock(
+            return_value=_fake_completion("What is the ideal gas law?")
+        )
+        with patch.object(expander, "_get_client", return_value=fake):
+            out = await expand_query(
+                "why?",
+                context_turns=[
+                    "student: what's PV=nRT used for?",
+                    "tutor: it relates pressure, volume, and temperature for gases.",
+                ],
+            )
+        assert out == "What is the ideal gas law?"
+        assert _outcome("expanded") == before + 1
+        # And the gateway actually got called
+        fake.chat.completions.create.assert_awaited_once()
+
+    @pytest.mark.asyncio
+    async def test_long_query_passthrough_no_gateway_call(self) -> None:
+        """Queries at/above the token threshold must skip expansion entirely."""
+        before = _outcome("passthrough_long")
+        fake = MagicMock()
+        fake.chat.completions.create = AsyncMock(
+            return_value=_fake_completion("should-not-appear")
+        )
+        with patch.object(expander, "_get_client", return_value=fake):
+            out = await expand_query(
+                "how does the ideal gas law relate pressure and volume",
+                context_turns=["irrelevant"],
+            )
+        assert out == "how does the ideal gas law relate pressure and volume"
+        assert _outcome("passthrough_long") == before + 1
+        fake.chat.completions.create.assert_not_awaited()
+
+    @pytest.mark.asyncio
+    async def test_no_context_passthrough(self) -> None:
+        """Short query with empty context must passthrough without a gateway call."""
+        before = _outcome("passthrough_nocontext")
+        fake = MagicMock()
+        fake.chat.completions.create = AsyncMock(
+            return_value=_fake_completion("should-not-appear")
+        )
+        with patch.object(expander, "_get_client", return_value=fake):
+            out = await expand_query("why?", context_turns=None)
+        assert out == "why?"
+        assert _outcome("passthrough_nocontext") == before + 1
+        fake.chat.completions.create.assert_not_awaited()
+
+    @pytest.mark.asyncio
+    async def test_blank_turns_treated_as_no_context(self) -> None:
+        """Whitespace-only turns must not trigger a gateway call."""
+        before = _outcome("passthrough_nocontext")
+        fake = MagicMock()
+        fake.chat.completions.create = AsyncMock()
+        with patch.object(expander, "_get_client", return_value=fake):
+            out = await expand_query("why?", context_turns=["   ", "", "\t\n"])
+        assert out == "why?"
+        assert _outcome("passthrough_nocontext") == before + 1
+        fake.chat.completions.create.assert_not_awaited()
+
+    @pytest.mark.asyncio
+    async def test_gateway_failure_falls_back_to_raw_query(self, caplog) -> None:
+        """Any exception from the gateway must produce raw query + WARNING."""
+        import logging as _logging
+
+        before = _outcome("fallback_error")
+        fake = MagicMock()
+        fake.chat.completions.create = AsyncMock(side_effect=RuntimeError("boom"))
+        with (
+            patch.object(expander, "_get_client", return_value=fake),
+            caplog.at_level(_logging.WARNING, logger="app.expander"),
+        ):
+            out = await expand_query("why?", context_turns=["tutor: gas laws"])
+        assert out == "why?"
+        assert _outcome("fallback_error") == before + 1
+        assert any("falling back" in rec.message for rec in caplog.records)
+
+    @pytest.mark.asyncio
+    async def test_blank_expansion_falls_back_to_raw_query(self) -> None:
+        """Whitespace-only gateway output must fall back, not propagate."""
+        before = _outcome("fallback_blank")
+        fake = MagicMock()
+        fake.chat.completions.create = AsyncMock(return_value=_fake_completion("   "))
+        with patch.object(expander, "_get_client", return_value=fake):
+            out = await expand_query("why?", context_turns=["tutor: gas laws"])
+        assert out == "why?"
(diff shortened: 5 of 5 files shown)


DreadPirateRobertz (commented):
## Code Review — tl-afb query expansion

**Reviewed:** `expander.py` (new), `routers/search.py`, `metrics.py`, `config.py`, `tests/test_expander.py`.

### Verdict: Approve with two non-blocking notes

The implementation matches the spec cleanly: total `expand_query` contract, lazy client, structured outcome counter with all five branches accounted for, graceful fallback on both exception and blank output (with `exc_info=True` on the error path — good for ops). `effective_q` is correctly threaded through `_run_search`, `rerank`, and the `SearchResponse.query` echo. Test suite covers all five outcome branches plus clipping and full endpoint wiring; the autouse `_reset_client` fixture is the right call to prevent state bleed.

### Confidence ≥ 80 issues
None.

### Notes (informational, < 80 confidence — not blocking)

1. **No explicit gateway timeout (~70).** `_get_client()` constructs `AsyncOpenAI` without a `timeout=`. A slow/hung gateway would stall the search request until the underlying httpx default fires. The `except Exception` correctly catches `asyncio.TimeoutError` *if* one is raised, but there's no upper bound configured here. If other call sites in the service set timeouts on gateway clients, mirror that pattern; otherwise consider a small `timeout=` (e.g. 2-3s) since this is a best-effort enhancement on the hot search path. Defer if this matches existing service convention.

2. **`context_turns` unbounded at API layer (~55).** FastAPI parses all repeated `context_turns=` params into memory before `_clip_turns` trims to `query_expansion_max_context_turns`. A pathological caller sending thousands of turns would allocate them all. Low practical risk behind auth, but a `max_items` or per-item length cap on the `Query(...)` would be a cheap hardening win.

### Things I checked and liked
- `passthrough_long` vs `passthrough_nocontext` vs `fallback_blank` vs `fallback_error` vs `expanded` — all five label values appear in both production code and tests, no orphans.
- `< query_expansion_short_threshold` matches the docstring wording ("fewer than"). Threshold 5 → ≤4 tokens expand.
- Whitespace-only turns correctly filtered before the "no context" check (see `test_blank_turns_treated_as_no_context`).
- Response `query` field deliberately echoes the rewritten form — documented in the router docstring. Downstream consumers logging the original user query need to know this is the *effective* query, not the raw input. Worth a one-liner in the bead's exit notes if not already there.
- Test for clipping verifies both that old turns are dropped and new turns are present — robust.
- `_get_client` is patchable at the module level; tests use `patch.object(expander, "_get_client", ...)` correctly rather than fighting the lazy init.

### CLAUDE.md compliance
- TDD: 9 tests for ~70 lines of logic — well above bar.
- Docstrings: every public function and the module itself has Google-style docstrings. Private helpers documented too.
- Codecov patch passing per CI.
- Branch name `feat/tl-afb-query-expansion` matches convention.

Ship it. Address the timeout + input-bound notes in a follow-up bead if the team agrees they're worth doing.

— Petra (code-reviewer)

DreadPirateRobertz (commented):
## Type Design Review — tl-afb query expansion

**Reviewer: Petra (type-design-analyzer)**

---

### Invariants Identified

- `expand_query` is **total**: it never raises, always returns a non-empty `str` (the original query on any failure path).
- `context_turns: list[str] | None` — `None` and `[]` are semantically distinct: `None` = caller provided no context at all; `[]` = explicitly empty list. Both are handled identically via `context_turns or []`, which is correct given the design intent.
- `_clip_turns` guarantees its output length is bounded by `query_expansion_max_context_turns` and contains only non-empty, stripped strings.
- `_is_short` is a pure predicate over config — no state, no side effects.
- Every expansion outcome increments exactly one `QUERY_EXPANSION_OUTCOMES` label before returning.

---

### Ratings

**Encapsulation: 8/10**
The lazy `_client` global is module-private and the tests patch `_get_client` rather than the global directly — that's the right seam. The only encapsulation concern is that `_client` is accessible as `expander._client` and the test fixture resets it manually. A `_reset_client()` module-level function would make the reset contract explicit rather than relying on attribute access to internals, but this is a minor point for a test-only concern.

**Invariant Expression: 9/10**
The docstring on `expand_query` states the total contract in plain language ("never raises", "drop-in replacement"). The return type `str` is clean and honest — there is no `str | None` escape hatch. `_is_short` and `_clip_turns` are narrow, well-named predicates that communicate exactly one responsibility each. The `Annotated[list[str] | None, Query(...)]` for the FastAPI param is idiomatic and the description string explains the wire format (`&context_turns=...` repeated). Invariant expression is strong throughout.

**Invariant Usefulness: 9/10**
The invariants directly prevent two real bug classes: (1) a gateway failure causing a 500 on search, and (2) prompt payload blowup from unbounded context. The `None` vs `[]` distinction is correctly preserved at the API boundary even though the implementation collapses them — this matters for future observability (you could log the `None` case differently) and avoids forcing the caller to always supply an empty list. The five outcome labels give operational visibility that would otherwise require log parsing.

**Invariant Enforcement: 8/10**
Construction-time enforcement is strong: `_clip_turns` strips and filters immediately, so `expand_query` never sees raw whitespace turns. The `try/except Exception` catch-all in `expand_query` is intentionally broad (total contract requires it) and the blank-output check covers the one case exceptions cannot. One gap: the `_is_short` threshold is a runtime config read, not a type-level constraint — a query of exactly `query_expansion_short_threshold` tokens goes to passthrough (`<`, not `<=`), which is correct but not expressed in the type. A comment on the predicate noting "strictly less than" would prevent an off-by-one correction from accidentally flipping the boundary in the future.

---

### Strengths

- `Iterable[str]` for `_clip_turns` is the right call. The function only needs iteration and slicing; accepting `Iterable[str]` documents that cleanly and makes the helper independently testable with any sequence. Since `_clip_turns` materializes to `list[str]` internally and returns `list[str]`, callers get a concrete type back regardless of what they pass in — no type-level confusion propagates outward.
- The decision not to introduce a `TypedDict` or `Protocol` is correct. The turns are plain strings with no structure that types can enforce; a wrapper would add ceremony with no invariant benefit.
- Return type `str` on `expand_query` is precise. Returning `str | None` to signal "no expansion happened" would have infected every call site with an unwrap. The total contract eliminates that footgun.
- The metric increment happens on every exit path before `return` — coverage is enforced by the test suite's counter assertions per outcome label, which is the right way to test this class of invariant.
- `test_endpoint_never_500s_on_expansion_failure` is valuable: it documents that the endpoint's safety depends on `expand_query` being total, not on a router-level try/except. That makes the design contract auditable.

---

### Concerns

1. **`_client` global reset in tests via attribute access.** `expander._client = None` in the fixture reaches into a private implementation detail. Low severity — tests pass — but if `_client` is ever renamed or moved into a class, the fixture silently stops resetting. Consider a `_reset_for_testing()` function or using `importlib.reload` if that fits the test infra.

2. **`_is_short` reads `settings` at call time.** This is fine for production but means the threshold is not a type-level constraint — any integer can be stored. If `query_expansion_short_threshold` were typed as `Annotated[int, Field(ge=1)]` in `Settings`, Pydantic would enforce positivity at startup rather than silently allowing a zero threshold that makes all queries "long." Worth a one-line addition to `config.py`.

3. **No explicit type for the `outcome` label values.** The five strings (`"expanded"`, `"passthrough_long"`, etc.) appear in both `expander.py` and the test's `_outcome()` helper as bare string literals. A `Literal["expanded", "passthrough_long", "passthrough_nocontext", "fallback_blank", "fallback_error"]` type alias in `metrics.py` would give mypy coverage and make the valid set self-documenting. This is purely additive.

---

### Recommended Improvements

1. In `config.py`, add `ge=1` to the threshold fields:
   ```python
   query_expansion_short_threshold: Annotated[int, Field(ge=1)] = 5
   query_expansion_max_context_turns: Annotated[int, Field(ge=1)] = 6
   ```

2. In `metrics.py`, add a type alias for outcome labels:
   ```python
   ExpansionOutcome = Literal[
       "expanded", "passthrough_long", "passthrough_nocontext",
       "fallback_blank", "fallback_error",
   ]
   ```
   Then use `ExpansionOutcome` in `expander.py` at the `labels(outcome=...)` call sites.

3. Add a one-word comment to `_is_short` clarifying the boundary: `# strictly less than threshold — == is treated as "long"`. Prevents a future "fix" from flipping `<` to `<=`.

None of these are blockers. The core type design is sound, the total contract is properly enforced and tested, and the encapsulation is good. The improvements are all additive and low-cost.

**Overall: approve with optional nits above.**

— Petra (type-design-analyzer)

DreadPirateRobertz (commented):
## Test Coverage Review — `services/search/tests/test_expander.py`

Reviewed: `app/expander.py`, `app/routers/search.py`, `app/metrics.py`, `tests/test_expander.py`

Overall the suite is solid. The happy-path, all five counter outcomes, turn-clipping, endpoint wiring, and graceful-failure contract are all covered. The delta-based counter pattern (`before = _outcome(...); assert _outcome(...) == before + 1`) is a clean way to avoid coupling to absolute counter values across a shared global registry. Four gaps worth addressing, one of which is important enough to block.

---

### Critical Gaps

**1. `choices=None` / empty list raises inside the `try` block — no dedicated test (criticality: 8)**

In `expander.py`:
```python
expanded = (response.choices[0].message.content or "").strip()
```
If the gateway returns a well-formed HTTP 200 but with `choices=None` or `choices=[]`, this raises `TypeError`/`IndexError` after the `await` completes. That exception is still inside the `try/except Exception` block, so it falls back correctly — but no test verifies that path. The existing `test_gateway_failure_falls_back_to_raw_query` only covers a `RuntimeError` raised by `create` itself, not a structurally malformed response. A future refactor that moves the unpacking outside the `try` (e.g. to validate the response shape) would silently break this contract.

Suggested addition (unit, inside `TestExpandQuery`):
```python
@pytest.mark.asyncio
async def test_choices_none_falls_back_to_raw_query(self) -> None:
    """Malformed response with choices=None must produce fallback_error, not crash."""
    before = _outcome("fallback_error")
    fake = MagicMock()
    # Gateway returns successfully but with no choices — triggers IndexError/TypeError
    fake.chat.completions.create = AsyncMock(
        return_value=SimpleNamespace(choices=None)
    )
    with patch.object(expander, "_get_client", return_value=fake):
        out = await expand_query("why?", context_turns=["tutor: gas laws"])
    assert out == "why?"
    assert _outcome("fallback_error") == before + 1
```
Also worth adding `choices=[]` as a second parametrized case. Both are real gateway response shapes seen when a model is overloaded or rate-limited.

---

### Important Improvements

**2. `_clip_turns` boundary: exactly `max_turns` entries is untested (criticality: 5)**

`test_context_is_clipped_to_max` uses `max_turns + 4` entries and checks that the oldest four were dropped. There is no test for the boundary cases:
- Exactly `max_turns` entries — slice `[-max_turns:]` on a list of length `max_turns` should return the full list unchanged.
- Exactly `max_turns - 1` entries — no clipping, all preserved.

The implementation `cleaned[-max_turns:]` is correct for both, but the boundary is the most likely place a future edit (e.g. switching to `[:max_turns]` by mistake) introduces a bug without failing any existing test. A one-liner parametrized test on `_clip_turns` directly would pin this:

```python
from app.expander import _clip_turns
from app.config import settings

def test_clip_turns_at_exact_max_preserves_all():
    turns = [f"t{i}" for i in range(settings.query_expansion_max_context_turns)]
    assert _clip_turns(turns) == turns

def test_clip_turns_below_max_preserves_all():
    turns = [f"t{i}" for i in range(settings.query_expansion_max_context_turns - 1)]
    assert _clip_turns(turns) == turns
```

**3. `stream=False` is never asserted in any gateway call (criticality: 6)**

The happy-path test verifies the gateway was `awaited_once()`, and `test_context_is_clipped_to_max` captures `**kwargs` to inspect `messages` — but neither test asserts `stream=False` in the captured kwargs. If `stream=False` is accidentally removed, the gateway call becomes a streaming response, `response.choices` returns an iterator instead of a list, the unpacking raises `TypeError` inside the `try` block, and every expansion silently falls back. No test turns red.

Add one assertion to `test_context_is_clipped_to_max` (it already captures `**kwargs`) or to the happy-path test:

```python
# In test_context_is_clipped_to_max, after the existing assertions:
assert captured.get("stream") is False
```

Or equivalently in the happy-path, promote from `assert_awaited_once()` to:
```python
fake.chat.completions.create.assert_awaited_once_with(
    model=ANY,
    messages=ANY,
    max_tokens=ANY,
    stream=False,
)
```

---

### Test Quality Notes

**4. Prometheus counter isolation is serial-safe but fragile to parallelization (criticality: 3)**

The delta pattern is correct in serial pytest runs. If the suite is ever run with `pytest-xdist -n auto` the global `REGISTRY` is shared across workers only if tests share a process — xdist forks, so in practice this is safe. The real fragility is if a future test helper or fixture calls `expand_query` as a side effect (e.g. a conftest startup fixture that warms the client), which would shift the counter before the `before =` snapshot. This is low risk now but worth a comment in the test file near `_outcome()` stating that the delta pattern assumes no concurrent or side-effectful counter increments within a single test session. No code change needed; a one-line comment is sufficient.

**5. `test_endpoint_never_500s_on_expansion_failure` documents a design assumption that could become false (criticality: 4)**

The test's docstring says: "the router itself is not wrapped in try/except, so this test documents that the expander contract is what protects the endpoint." That is accurate today. But if someone later wraps the router in a try/except that swallows expander errors and returns 200 with an empty body, this test still passes — it only asserts `status_code == 200` and `query == "why?"`, which are also met by a broken fallback. Consider adding an assertion that `results` is a list (even if empty) to ensure the pipeline completed, not just that the status was 200.

---

### Positive Observations

- All five counter outcomes (`expanded`, `passthrough_long`, `passthrough_nocontext`, `fallback_error`, `fallback_blank`) have dedicated tests — the Prometheus coverage is comprehensive.
- The `autouse` fixture resetting `expander._client = None` is the right approach to prevent lazy-init leakage between tests.
- `test_short_query_with_context_is_expanded_in_pipeline` correctly captures both the embedder and rerank call arguments, verifying the effective query flows through the entire pipeline — not just the response body.
- `test_blank_turns_treated_as_no_context` correctly distinguishes `""`, `" "`, and `"\t\n"` — covers the filter in `_clip_turns` that strips before checking.
- The `_fake_completion` helper correctly mirrors the OpenAI SDK's nested `SimpleNamespace` shape, making the mocks representative.

---

**Summary**: One gap should be closed before merge (the `choices=None`/`choices=[]` path — criticality 8). The `stream=False` assertion (6) and `_clip_turns` boundary tests (5) are recommended follow-ups. All other items are minor. Overall this is well-tested work for a new async module with five distinct counter outcomes and graceful-failure requirements.

Petra (pr-test-analyzer)

DreadPirateRobertz (commented):
## Comment Analysis — PR #239 (tl-afb query expansion)

Petra (comment-analyzer)

---

### Summary

Five files reviewed: `expander.py` (new), `config.py`, `metrics.py`, `search.py` (router), `test_expander.py`. The documentation is largely accurate and well-structured. Three items need attention: one inaccurate word in the router docstring, one ambiguous phrase in `_clip_turns`, and three magic numbers in `config.py` that are under-explained.

---

### Critical Issues

**Location: `services/search/app/routers/search.py`, router docstring addition**

> "Expansion failures silently fall back to the raw query (tl-afb)."

The word "silently" is factually wrong. Both fallback paths (`fallback_error` and `fallback_blank`) emit a `logger.warning(...)` call **and** increment the `search_query_expansion_total` metric counter. Neither path is silent. A future maintainer reading this docstring will expect no observable signal on failure — that is the opposite of what the code does, and it is also the opposite of the spec intent documented in the module docstring ("log a WARNING, and increment the counter").

Suggested fix:

```
When ``q`` is a short follow-up (< ``query_expansion_short_threshold``
tokens) and ``context_turns`` is non-empty, the query is first rewritten
into a standalone form via the AI gateway. Expansion failures fall back
to the raw query with a WARNING log and a ``search_query_expansion_total``
counter increment (tl-afb).
```

---

### Improvement Opportunities

**Location: `services/search/app/expander.py`, `_clip_turns` docstring**

> "Keep the most recent N non-empty turns, bounded by config."

The actual filter is `t and t.strip()`, which removes both zero-length strings and whitespace-only strings (spaces, tabs, newlines). The word "non-empty" in English typically means only the zero-length case. A maintainer seeing a turn of `"   "` might not immediately realize it is also discarded. The test `test_blank_turns_treated_as_no_context` covers this behavior, but the docstring does not.

Suggested fix:

```
Keep the most recent N turns that are non-empty after stripping whitespace,
bounded by ``query_expansion_max_context_turns``.
```

**Location: `services/search/app/config.py`, query expansion block comment**

```python
# Query expansion via AI Gateway (tl-afb) — used for short (<5-token)
# follow-up queries when the caller supplies conversation context turns.
tutor_fast_model: str = "tutor-fast"
query_expansion_short_threshold: int = 5
query_expansion_max_context_turns: int = 6
query_expansion_max_tokens: int = 128
```

The comment explains *what* the block is for but does not explain *why* the three numeric defaults were chosen. All three will look like arbitrary magic numbers to a future maintainer tuning the service:

- `5` tokens: what is the reasoning? Is this the point at which queries empirically become underspecified? The comment says `<5-token` which gives the value but not the rationale.
- `6` turns: why 6 and not 4 or 8? Is this a cost/context tradeoff? A round-trip typical conversation length?
- `128` max_tokens: what happens if the AI gateway returns a truncated (128-token) rewrite? Is that silently used as-is? The blank-output fallback handles empty output but not a mid-sentence truncation.

Suggested addition (inline or as an extended comment above the block):

```python
# Query expansion via AI Gateway (tl-afb).
# Triggers when: query has fewer than query_expansion_short_threshold whitespace
# tokens AND caller supplies context_turns. Defaults chosen per tl-afb spec:
#   short_threshold=5  — empirically, ≤4-token queries are almost always
#                        underspecified follow-ups ("why?", "and that one?")
#   max_context_turns=6 — keeps gateway payload small; 6 turns covers ~3
#                         student/tutor exchanges, sufficient for disambiguation
#   max_tokens=128     — a self-contained query rarely exceeds ~30 tokens;
#                        128 is a safe ceiling. Truncated rewrites are used
#                        as-is (no additional validation).
```

(Fill in the actual rationale from the spec or design discussion — the placeholders above are illustrative.)

---

### Recommended Removals

None. Every comment earns its place.

---

### Positive Findings

**`expand_query` docstring** — The "Returns" section explicitly documents the total/never-raises contract and phrases it as a caller-facing guarantee ("callers can treat its output as a drop-in replacement"). This is exactly the right level of documentation for a function with a non-obvious behavioral contract.

**Metrics comment block** — All five `outcome` label values are enumerated inline with their precise semantics. This is the correct pattern: Prometheus label values are stringly typed and invisible to IDEs, so enumerating them next to the `Counter` definition is the only reliable discoverability mechanism.

**`test_endpoint_never_500s_on_expansion_failure` inline comment** — The comment explicitly calls out that the router has no try/except and that the expander contract is the sole protection for the endpoint. This is accurate, self-aware, and flags a risk that a future refactor should not break. Good defensive documentation.

**`_get_client()` placement inside try** — The question raised in the PR brief was whether initialization errors are caught. Confirmed: `client = _get_client()` is the first line inside the `try:` block, so `AsyncOpenAI(...)` constructor failures are caught by the bare `except Exception:`. The "never raises" contract holds.

---

Petra (comment-analyzer)

DreadPirateRobertz (commented):
## Error Handling Review — expander.py (tl-afb)

Reviewed as silent-failure-hunter. Overall the structure is solid: `exc_info=True` is present, the fallback contract is stated, counters increment on every path. But there are several issues that would make debugging genuinely painful in production.

---

### CRITICAL — ImportError swallowed by `except Exception`

**Location:** `services/search/app/expander.py`, `_get_client()` + the try block

`_get_client()` does a deferred `from openai import AsyncOpenAI` inside the function body. Because `_get_client()` is called *inside* the try block, an `ImportError` (missing openai package, broken venv, bad install) gets caught by `except Exception`, increments `fallback_error`, and logs "query expansion failed — falling back to raw query."

A deployment engineer staring at that WARNING has zero indication that the openai package is not installed. This is a silent misconfiguration that would manifest as every short query silently degrading to unexpanded results, with a log line that looks like a transient gateway hiccup. It could go unnoticed for days.

The deferred import pattern is fine for testability, but the import should be guarded separately:

```python
def _get_client():
    try:
        from openai import AsyncOpenAI
    except ImportError as exc:
        raise RuntimeError(
            "openai package is required for query expansion (tl-afb); "
            "install it or set QUERY_EXPANSION_ENABLED=false"
        ) from exc
    global _client
    if _client is None:
        _client = AsyncOpenAI(
            base_url=settings.ai_gateway_url,
            api_key=settings.openai_api_key or "ai-gateway-local",
        )
    return _client
```

An `ImportError` at startup/first-call is a hard deployment fault, not a graceful fallback scenario. It should propagate and crash loudly — not hide behind `fallback_error`.

---

### HIGH — No timeout on the gateway completion call

**Location:** `services/search/app/expander.py` lines ~100-107

```python
response = await client.chat.completions.create(
    model=settings.tutor_fast_model,
    messages=prompt,
    max_tokens=settings.query_expansion_max_tokens,
    stream=False,
)
```

No `timeout` parameter. The AsyncOpenAI client defaults to 600 seconds. This call is in the hot path of every short search query. A hung or slow gateway will stall those requests for up to 10 minutes before the `except Exception` fires.

The spec explicitly calls this feature "cheap." Enforce that at the network layer:

```python
response = await client.chat.completions.create(
    model=settings.tutor_fast_model,
    messages=prompt,
    max_tokens=settings.query_expansion_max_tokens,
    stream=False,
    timeout=settings.query_expansion_timeout_seconds,  # add to config, suggest 3.0
)
```

Add `query_expansion_timeout_seconds: float = 3.0` to `config.py`. A 3-second budget is generous for a 128-token completion; anything beyond that should count as a failure and fall back immediately.

---

### HIGH — `fallback_error` conflates structurally different failure modes

**Location:** `services/search/app/expander.py`, the `except Exception` block

`response.choices[0]` raises `IndexError` when the gateway returns an empty choices list (HTTP 200, valid JSON, zero completions). This is a gateway misconfiguration or model-level rejection — structurally different from a network error, auth failure, or timeout, all of which also land in `fallback_error`.

In production, if the gateway starts returning empty choices consistently, the `fallback_error` counter will spike — but the log message "query expansion failed" with an `IndexError` traceback buried in `exc_info` is not going to prompt the right remediation. Someone will look for network issues when the real problem is a model routing misconfiguration.

Suggested approach: check `response.choices` length explicitly before indexing, and give it a distinct outcome:

```python
if not response.choices:
    QUERY_EXPANSION_OUTCOMES.labels(outcome="fallback_empty_choices").inc()
    logger.warning(
        "query expansion: gateway returned empty choices list "
        "(model=%s) — falling back to raw query",
        settings.tutor_fast_model,
    )
    return query
expanded = (response.choices[0].message.content or "").strip()
```

This also eliminates the `IndexError` path from `except Exception` entirely, making the catch block narrower.

---

### HIGH — Fallback warning log carries no diagnostic context

**Location:** `services/search/app/expander.py` lines ~103-106

```python
logger.warning(
    "query expansion failed — falling back to raw query",
    exc_info=True,
)
```

`exc_info=True` is good — the traceback is there. But the log carries none of:
- Which query triggered the failure
- Which model was targeted (`settings.tutor_fast_model`)
- Which gateway URL was called (`settings.ai_gateway_url`)
- How many context turns were supplied

Six months from now, a Sentry alert on this warning gives you a traceback and nothing else to correlate it with. Suggested:

```python
logger.warning(
    "query expansion failed — falling back to raw query "
    "(model=%s, gateway=%s, query_len=%d, context_turns=%d)",
    settings.tutor_fast_model,
    settings.ai_gateway_url,
    len(query),
    len(clipped),
    exc_info=True,
)
```

Same fix applies to the `fallback_blank` path — it currently logs zero context. At minimum log the model and that you got a blank, so an operator can correlate with gateway logs.

---

### MEDIUM — `except Exception` can catch bugs in prompt-building code

**Location:** `services/search/app/expander.py`, the try block

The try block wraps not just the `await` but also `client = _get_client()`. If a future developer adds logic between `_get_client()` and the `await` (e.g. prompt manipulation, logging, preprocessing), any bug in that code — a `KeyError`, `AttributeError`, `NameError` — will be caught by `except Exception` and silently attributed to "query expansion failed." The `exc_info=True` saves you here, but the counter label `fallback_error` will be misleading.

The try block should wrap only the I/O operation:

```python
client = _get_client()  # outside try — let this propagate on hard failure
try:
    response = await client.chat.completions.create(...)
    if not response.choices:
        ...
    expanded = (response.choices[0].message.content or "").strip()
except (openai.APIError, openai.APIConnectionError, openai.RateLimitError,
        openai.APITimeoutError, asyncio.TimeoutError, IndexError, AttributeError):
    QUERY_EXPANSION_OUTCOMES.labels(outcome="fallback_error").inc()
    logger.warning(...)
    return query
```

Catching specific openai exception types also makes the catch self-documenting about what failure modes were anticipated.

---

### MEDIUM — Race condition in `_get_client()` lazy init

**Location:** `services/search/app/expander.py` lines ~35-43

```python
_client = None

def _get_client():
    global _client
    if _client is None:
        ...
        _client = AsyncOpenAI(...)
    return _client
```

Two concurrent coroutines can both observe `_client is None`, both construct an `AsyncOpenAI` instance, and the second write wins. The first client is leaked. In CPython this is unlikely to cause functional problems (both clients are identical), but it's a correctness issue. Use `asyncio.Lock` or module-level initialization:

```python
_client: AsyncOpenAI | None = None
_client_lock = asyncio.Lock()

async def _get_client() -> AsyncOpenAI:
    global _client
    async with _client_lock:
        if _client is None:
            ...
            _client = AsyncOpenAI(...)
    return _client
```

Or initialize at import time (after settings are available) rather than lazily — the deferred import for testability can still be done, just eagerly on first module load.

---

### LOW — Router docstring says "silently fall back" — misleading language

**Location:** `services/search/app/routers/search.py`, the `search()` docstring addition

> "Expansion failures silently fall back to the raw query (tl-afb)."

They don't fail silently — they log WARNING and increment a counter. "Silently" in error-handling contexts means "without any observable signal," which is precisely what this PR avoids. This will confuse the next developer who reads it. Suggest: "Expansion failures are logged as WARNING and fall back to the raw query."

---

### LOW — `test_endpoint_never_500s_on_expansion_failure` does not test its stated guarantee

**Location:** `services/search/tests/test_expander.py`, `TestSearchEndpointWiring`

The test comment says "if a future refactor breaks that contract, the endpoint should still return 200" — but the test relies entirely on `expander.py`'s own try/except to pass. If the try/except were removed, this test would fail with a 500. The test does not isolate endpoint-level resilience from expander-level resilience. It documents the current behavior but does not enforce the stated guarantee. Not blocking, but worth acknowledging.

---

## Summary

| Severity | Count | Key items |
|----------|-------|-----------|
| CRITICAL | 1 | ImportError swallowed — missing openai package looks like a transient gateway hiccup |
| HIGH | 3 | No gateway timeout; empty-choices IndexError conflated with network errors; fallback warning has no diagnostic context |
| MEDIUM | 2 | Broad except catches future bugs in non-I/O code; lazy-init race |
| LOW | 2 | Misleading "silently" in docstring; test overstates its guarantee |

The CRITICAL (ImportError) and the missing timeout are the two I'd require fixed before merge — both could cause extended silent degradation in production. The diagnostic context in the warning log is a one-liner fix that has outsized value for on-call debugging.

— Petra (silent-failure-hunter)

DreadPirateRobertz (commented):
## Test Coverage Review — services/search/tests/test_expander.py

Reviewed: app/expander.py, app/routers/search.py, app/metrics.py, tests/test_expander.py

Overall the suite is solid. The happy-path, all five counter outcomes, turn-clipping, endpoint wiring, and graceful-failure contract are all covered. The delta-based counter pattern (before = _outcome(...); assert _outcome(...) == before + 1) is a clean way to avoid coupling to absolute counter values across a shared global registry. Four gaps worth addressing, one of which is important enough to block.

---

### Critical Gaps

**1. choices=None / empty list raises inside the try block — no dedicated test (criticality: 8)**

In expander.py:

    expanded = (response.choices[0].message.content or "").strip()

If the gateway returns a well-formed HTTP 200 but with choices=None or choices=[], this raises TypeError/IndexError after the await completes. That exception is still inside the try/except Exception block, so the code falls back correctly — but no test verifies this path. The existing test_gateway_failure_falls_back_to_raw_query only covers a RuntimeError raised by create itself, not a structurally malformed response object. A future refactor that moves the unpacking outside the try (e.g. to validate the response shape first) would silently break this contract with no failing test.

Suggested addition (unit, inside TestExpandQuery):

    @pytest.mark.asyncio
    async def test_choices_none_falls_back_to_raw_query(self) -> None:
        before = _outcome("fallback_error")
        fake = MagicMock()
        fake.chat.completions.create = AsyncMock(
            return_value=SimpleNamespace(choices=None)
        )
        with patch.object(expander, "_get_client", return_value=fake):
            out = await expand_query("why?", context_turns=["tutor: gas laws"])
        assert out == "why?"
        assert _outcome("fallback_error") == before + 1

Also add choices=[] as a second parametrized case. Both are real gateway response shapes seen when a model is overloaded or rate-limited.

---

### Important Improvements

**2. _clip_turns boundary: exactly max_turns entries is untested (criticality: 5)**

test_context_is_clipped_to_max uses max_turns + 4 entries and checks that the oldest four were dropped. There is no test for:
- Exactly max_turns entries — slice [-max_turns:] on a list of length max_turns should return the full list unchanged.
- Exactly max_turns - 1 entries — no clipping, all preserved.

The implementation is correct for both, but the boundary is the most likely place a future edit (e.g. switching from [-max_turns:] to [:max_turns] by mistake) introduces a bug without failing any existing test. Two simple unit tests on _clip_turns directly would pin this:

    def test_clip_turns_at_exact_max_preserves_all():
        turns = [f"t{i}" for i in range(settings.query_expansion_max_context_turns)]
        assert _clip_turns(turns) == turns

    def test_clip_turns_below_max_preserves_all():
        turns = [f"t{i}" for i in range(settings.query_expansion_max_context_turns - 1)]
        assert _clip_turns(turns) == turns

**3. stream=False is never asserted in any gateway call (criticality: 6)**

The happy-path test calls assert_awaited_once() and test_context_is_clipped_to_max captures **kwargs to inspect messages — but neither asserts stream=False in the captured kwargs. If stream=False is accidentally removed, the gateway returns an async iterator instead of a response object, response.choices[0] raises, and every expansion silently falls back to the raw query. No test turns red.

The fix is one assertion in test_context_is_clipped_to_max (which already captures kwargs):

    assert captured.get("stream") is False

Or promote the happy-path assertion to:

    fake.chat.completions.create.assert_awaited_once_with(
        model=ANY, messages=ANY, max_tokens=ANY, stream=False
    )

---

### Test Quality Notes

**4. Prometheus counter isolation is serial-safe but fragile to parallelization (criticality: 3)**

The delta pattern is correct in serial pytest runs. The real fragility is if a future conftest fixture calls expand_query as a side effect (e.g. a startup warm-up), shifting the counter before the before = snapshot. This is low risk now. A comment near _outcome() stating the delta pattern assumes no concurrent or side-effectful increments within a single test session would make this explicit. No code change required.

**5. test_endpoint_never_500s_on_expansion_failure documents a design assumption that could silently pass even if broken (criticality: 4)**

The test only asserts status_code == 200 and query == "why?". If a future router change catches and swallows expander errors and returns 200 with an empty or broken body, this test still passes. Adding an assertion that results is a list (even if empty) would confirm the pipeline completed, not just that the status was non-500.

---

### Positive Observations

- All five counter outcomes (expanded, passthrough_long, passthrough_nocontext, fallback_error, fallback_blank) have dedicated tests. The Prometheus coverage is comprehensive.
- The autouse fixture resetting expander._client = None correctly prevents lazy-init state leakage between tests.
- test_short_query_with_context_is_expanded_in_pipeline captures both the embedder and rerank arguments, verifying the effective query flows through the entire pipeline and not just the response body.
- test_blank_turns_treated_as_no_context correctly distinguishes "", " ", and "\t\n" — covers the strip-before-check filter in _clip_turns.
- _fake_completion correctly mirrors the OpenAI SDK nested SimpleNamespace shape, making the mocks representative of real responses.

---

Summary: One gap should be closed before merge (choices=None/choices=[] path, criticality 8). The stream=False assertion (6) and _clip_turns boundary tests (5) are recommended follow-ups that can land in a fast-follow. All other items are minor. Overall this is well-tested work for a new async module with five distinct counter outcomes and an explicit graceful-failure contract.

Petra (pr-test-analyzer)

DreadPirateRobertz (commented):
## Test Coverage Review — services/search/tests/test_expander.py

Reviewed: app/expander.py, app/routers/search.py, app/metrics.py, tests/test_expander.py

Overall the suite is solid. The happy-path, all five counter outcomes, turn-clipping, endpoint wiring, and graceful-failure contract are all covered. The delta-based counter pattern (before = _outcome(...); assert _outcome(...) == before + 1) is a clean way to avoid coupling to absolute counter values across a shared global registry. Four gaps worth addressing, one of which is important enough to block.

---

### Critical Gaps

**1. choices=None / empty list raises inside the try block — no dedicated test (criticality: 8)**

In expander.py:

```python
expanded = (response.choices[0].message.content or "").strip()
```

If the gateway returns a well-formed HTTP 200 but with `choices=None` or `choices=[]`, this raises `TypeError`/`IndexError` after the await completes. That exception is still inside the `try/except Exception` block, so the code falls back correctly — but no test verifies this path. The existing `test_gateway_failure_falls_back_to_raw_query` only covers a `RuntimeError` raised by `create` itself, not a structurally malformed response object. A future refactor that moves the unpacking outside the `try` (e.g. to validate the response shape first) would silently break this contract with no failing test.

Suggested addition (unit, inside `TestExpandQuery`):

```python
@pytest.mark.asyncio
async def test_choices_none_falls_back_to_raw_query(self) -> None:
    before = _outcome("fallback_error")
    fake = MagicMock()
    fake.chat.completions.create = AsyncMock(
        return_value=SimpleNamespace(choices=None)
    )
    with patch.object(expander, "_get_client", return_value=fake):
        out = await expand_query("why?", context_turns=["tutor: gas laws"])
    assert out == "why?"
    assert _outcome("fallback_error") == before + 1
```

Also add `choices=[]` as a second parametrized case. Both are real gateway response shapes seen when a model is overloaded or rate-limited.

---

### Important Improvements

**2. _clip_turns boundary: exactly max_turns entries is untested (criticality: 5)**

`test_context_is_clipped_to_max` uses `max_turns + 4` entries and verifies the oldest four were dropped. There is no test for:
- Exactly `max_turns` entries — slice `[-max_turns:]` on a list of length `max_turns` returns the full list unchanged.
- Exactly `max_turns - 1` entries — no clipping, all preserved.

The implementation is correct for both, but this boundary is the most likely place a future edit (switching from `[-max_turns:]` to `[:max_turns]` by mistake) introduces a bug with no failing test. Two simple direct tests on `_clip_turns` would pin it:

```python
def test_clip_turns_at_exact_max_preserves_all():
    turns = [f"t{i}" for i in range(settings.query_expansion_max_context_turns)]
    assert _clip_turns(turns) == turns

def test_clip_turns_below_max_preserves_all():
    turns = [f"t{i}" for i in range(settings.query_expansion_max_context_turns - 1)]
    assert _clip_turns(turns) == turns
```

**3. stream=False is never asserted in any gateway call (criticality: 6)**

The happy-path test calls `assert_awaited_once()` and `test_context_is_clipped_to_max` captures `**kwargs` to inspect `messages` — but neither asserts `stream=False` in the captured kwargs. If `stream=False` is accidentally removed, the gateway returns an async iterator instead of a response object, `response.choices[0]` raises, and every expansion silently falls back to the raw query. No test turns red.

Fix: one assertion in `test_context_is_clipped_to_max` (which already captures kwargs):

```python
assert captured.get("stream") is False
```

Or promote the happy-path assertion:

```python
fake.chat.completions.create.assert_awaited_once_with(
    model=ANY, messages=ANY, max_tokens=ANY, stream=False
)
```

---

### Test Quality Notes

**4. Prometheus counter isolation is serial-safe but fragile to future parallelization (criticality: 3)**

The delta pattern is correct in serial pytest runs. The risk is if a future conftest fixture calls `expand_query` as a side effect (e.g. a startup warm-up), shifting the counter before the `before =` snapshot. A comment near `_outcome()` stating the delta pattern assumes no concurrent or side-effectful increments within a single session would make this explicit. No code change required.

**5. test_endpoint_never_500s_on_expansion_failure could pass even if the pipeline is broken (criticality: 4)**

The test only asserts `status_code == 200` and `query == "why?"`. If a future router change catches and swallows expander errors and returns 200 with an empty or malformed body, this test still passes. Adding an assertion that `results` is a list (even if empty) would confirm the pipeline completed, not just that the status was non-500.

---

### Positive Observations

- All five counter outcomes (expanded, passthrough_long, passthrough_nocontext, fallback_error, fallback_blank) have dedicated tests. Prometheus coverage is comprehensive.
- The `autouse` fixture resetting `expander._client = None` correctly prevents lazy-init state leakage between tests.
- `test_short_query_with_context_is_expanded_in_pipeline` captures both embedder and rerank arguments, verifying the effective query flows through the entire pipeline — not just the response body.
- `test_blank_turns_treated_as_no_context` correctly distinguishes `""`, `" "`, and `"\t\n"` — covers the strip-before-check filter in `_clip_turns`.
- `_fake_completion` correctly mirrors the OpenAI SDK nested `SimpleNamespace` shape, making mocks representative of real responses.

---

**Summary**: One gap should be closed before merge (choices=None/choices=[] path, criticality 8). The stream=False assertion (criticality 6) and _clip_turns boundary tests (criticality 5) are recommended follow-ups. All other items are minor. Overall this is well-tested work for a new async module with five distinct counter outcomes and an explicit graceful-failure contract.

Petra (pr-test-analyzer)

DreadPirateRobertz (commented):
## 5-Agent Review — PR #239 feat(tl-afb): query expansion

Agents run: code-reviewer · silent-failure-hunter · type-design-analyzer · comment-analyzer · pr-test-analyzer

---

### BLOCKING — must fix before merge

**[1] `config.py` — missing `Field` validators on all 4 new settings** *(type-design-analyzer)*

A misconfigured deployment silently disables or permanently breaks the feature with no startup error. `query_expansion_short_threshold=0` makes every query hit `passthrough_long`; `query_expansion_timeout_seconds=0` causes every gateway call to timeout immediately, surfacing as constant `fallback_error` indistinguishable from a real outage.

```python
from pydantic import Field

query_expansion_short_threshold: int    = Field(default=5,   ge=1)
query_expansion_max_context_turns: int  = Field(default=6,   ge=1)
query_expansion_max_tokens: int         = Field(default=128, ge=1, le=1024)
query_expansion_timeout_seconds: float  = Field(default=3.0, gt=0.0)
```

**[2] `expander.py:103-119` — single `try` block conflates init failures with gateway failures** *(silent-failure-hunter)*

`_get_client()` raising `ImportError` (missing package) or `ValidationError` (bad `ai_gateway_url`) is a deployment defect, not a transient gateway error. Both currently surface as identical `fallback_error` WARNING. Split into two scopes:

```python
try:
    client = _get_client()
except Exception:
    logger.error("query expansion client init failed — check ai_gateway_url and openai package", exc_info=True)
    QUERY_EXPANSION_OUTCOMES.labels(outcome="fallback_error").inc()
    return query

try:
    response = await client.chat.completions.create(...)
    expanded = (response.choices[0].message.content or "").strip()
except Exception:
    QUERY_EXPANSION_OUTCOMES.labels(outcome="fallback_error").inc()
    logger.warning("query expansion gateway call failed — falling back to raw query", exc_info=True)
    return query
```

**[3] `search.py:69` — no router-level guard on `expand_query`** *(silent-failure-hunter)*

If a future refactor legitimately narrows the `except` in `expand_query`, the router has no fallback and will 500. Add an independent guard:

```python
try:
    effective_q = await expand_query(q, context_turns)
except Exception:
    logger.error("expand_query raised unexpectedly — falling back to raw query", exc_info=True)
    effective_q = q
```

**[4] `test_expander.py` — missing threshold boundary tests** *(pr-test-analyzer)*

No test pins the `<` vs `<=` boundary at `query_expansion_short_threshold=5`. A change from `<` to `<=` passes all current tests silently.

```python
async def test_exactly_at_threshold_is_not_short(self) -> None:
    # 5 tokens == threshold → passthrough_long (not expanded)
    before = _outcome("passthrough_long")
    fake = MagicMock(); fake.chat.completions.create = AsyncMock(return_value=_fake_completion("nope"))
    with patch.object(expander, "_get_client", return_value=fake):
        out = await expand_query("one two three four five", context_turns=["ctx"])
    assert out == "one two three four five"
    assert _outcome("passthrough_long") == before + 1
    fake.chat.completions.create.assert_not_awaited()

async def test_one_below_threshold_is_short(self) -> None:
    # 4 tokens < threshold → expanded
    fake = MagicMock(); fake.chat.completions.create = AsyncMock(return_value=_fake_completion("expanded"))
    with patch.object(expander, "_get_client", return_value=fake):
        out = await expand_query("one two three four", context_turns=["ctx"])
    assert out == "expanded"
    fake.chat.completions.create.assert_awaited_once()
```

**[5] `test_expander.py` — missing `context_turns=[]` test** *(pr-test-analyzer)*

`test_no_context_passthrough` only covers `None`. FastAPI may pass `[]` when the param is absent in some versions. The `or []` guard is load-bearing and untested.

```python
async def test_empty_list_context_treated_as_no_context(self) -> None:
    before = _outcome("passthrough_nocontext")
    fake = MagicMock(); fake.chat.completions.create = AsyncMock()
    with patch.object(expander, "_get_client", return_value=fake):
        out = await expand_query("why?", context_turns=[])
    assert out == "why?"
    assert _outcome("passthrough_nocontext") == before + 1
    fake.chat.completions.create.assert_not_awaited()
```

**[6] `expander.py` — empty `query` string hits the gateway** *(type-design-analyzer)*

`expand_query("", context_turns=["ctx"])` passes `_is_short` (0 < 5) and reaches the gateway with a blank prompt. Add before the short-threshold check:

```python
if not query.strip():
    QUERY_EXPANSION_OUTCOMES.labels(outcome="passthrough_nocontext").inc()
    return query
```

**[7] `search.py` — `_run_search` docstring says "Raw query string" — now stale** *(comment-analyzer)*

This PR changes every `_run_search` call site to pass `effective_q` (post-expansion). Update the `Args` docstring:
```
q: Effective query string — raw or AI-gateway-expanded by the caller.
```

---

### Non-blocking (follow-up bead recommended)

- `expander.py:123` — `fallback_blank` WARNING has no structured context; add `extra={"raw_query": query, "turn_count": len(clipped)}` *(silent-failure-hunter — MEDIUM)*
- `metrics.py` — define `ExpansionOutcome = Literal[...]` to make counter label strings type-checkable *(type-design-analyzer)*
- `_clip_turns` — add hard DoS cap before deserialization (`[:50]`, per-turn `[:2000]`) *(type-design-analyzer)*
- `_clip_turns` docstring — "non-empty" → "non-blank (whitespace-only strings also discarded)" *(comment-analyzer)*
- `test_context_is_clipped_to_max` — add inline comment noting assertions assume `max_turns >= 4` *(comment-analyzer)*

---

**Code-reviewer:** No bugs found at confidence ≥ HIGH ✓

7 blocking fixes required. All sent to @dink directly. Re-request review after push.

PR #241, opened 2026-04-13
feat(tl-uzn): gaming-service WebSocket battle-state API

DreadPirateRobertz:
## What
**New Behavior**: A new endpoint \`GET /gaming/battle/{battleId}/ws\` upgrades to a WebSocket and streams live battle events (\`damage\`, \`phase_transition\`, \`loot_roll\`, plus a \`joined\` lifecycle frame) to the connected client until it disconnects. REST endpoints remain the only write path; the WS is broadcast-only.

**Previous Behavior**: The frontend had to poll \`GET /gaming/boss/session/{sessionId}\` to observe state changes.

## Detail
- New package \`internal/wsbattle\` — in-memory pub/sub \`Hub\` keyed by \`battle_id\`, concurrent-safe, non-blocking publish, unsubscribe closes the channel cleanly.
- New handler \`BattleWS\` — JWT validated from \`?token=\` (WS can't carry custom headers), ownership check against the stored battle session, ping/pong keep-alive, read-pump so disconnects are observed within the 60s deadline.
- \`Attack\`, \`ForfeitBattle\`, and \`finishBattle\` now publish to the hub after state is persisted — subscribers never see a phantom state that could be rolled back.
- Route registered outside the \`Authenticate\` middleware group in \`cmd/server/main.go\` (auth happens inside the handler instead).
- New dep: \`github.com/gorilla/websocket v1.5.3\`.

## Why
Bead: **tl-uzn** — backend counterpart to the tl-dye frontend polish. Real-time HP bar animation + loot reveal need a push channel; polling at 10+ Hz to catch a 500ms boss strike is wasteful.

## How to test
\`\`\`
cd services/gaming-service
go test -race -timeout 60s ./internal/wsbattle/... ./internal/handler/...
\`\`\`

Event shapes (consumers):
\`\`\`json
{"type":"damage","battle_id":"...","timestamp":"...","payload":{"player_damage_dealt":10,"boss_hp":190,"player_hp":100,"turn":1,"phase":"active","answer_correct":true}}
{"type":"phase_transition","battle_id":"...","payload":{"phase":"victory","player_hp":80,"boss_hp":0}}
{"type":"loot_roll","battle_id":"...","payload":{"xp_earned":500,"gems_earned":10,"achievement":{...},"cosmetic":{...},"quote":"..."}}
\`\`\`

## Checklist
- [x] **TDD**: \`hub_test.go\` (7 tests incl. concurrent -race stress) + \`battle_ws_test.go\` (8 tests: connect/broadcast/disconnect/auth-matrix + Attack→victory broadcast chain)
- [x] **Coverage**: \`wsbattle\` 95.3%, \`handler\` 90.6% (both above the 90% gate)
- [x] **Docstrings**: package + every exported type, function, method; inline comments on the tricky bits (lock scoping around non-blocking send, token-from-query rationale)
- [x] All tests pass with \`-race\` (423 tests across gaming-service)
- [x] \`go vet ./...\` clean
- [x] PR body includes bead ID, what changed, why, how to test
- [x] No secrets committed
- [x] Self-review: read my own diff before opening


DreadPirateRobertz:
## Code-Reviewer — BLOCKING (security): Algorithm confusion in parseBattleWSToken (confidence 95). The parseBattleWSToken keyfunc returns the secret with no HMAC method assertion — every other auth endpoint checks jwt.SigningMethodHMAC. Fix: mirror middleware/auth.go parseToken — add 'if _, ok := t.Method.(*jwt.SigningMethodHMAC); !ok { return nil, fmt.Errorf(unexpected signing method) }'. Must fix before merge. Also: TestBattleWS_BroadcastEventDelivered SubscriberCount poll is dead (guard always false — Subscribe fires before joined-frame write).

DreadPirateRobertz:
## Comment-Analyzer finding

**wsUpgrader comment contains false security claim** — battle_ws.go wsUpgrader block:

Current comment claims 'browsers can't set arbitrary Origin headers from untrusted JS' as justification for CheckOrigin: return true. This is WRONG — browsers DO allow cross-site WebSocket connections (CSWSH). The real protection is the JWT in the query param. Fix the comment:

'CheckOrigin is permissive because the JWT query parameter is the real authentication gate. A valid token cannot be obtained cross-site without the user's credentials, so CSWSH is not viable. If this service is ever exposed without token validation, origin checking must be reinstated.'

Also: parseBattleWSToken godoc says 'validated the same way middleware.Authenticate does' — inaccurate since it lacks the HMAC algorithm assertion. Fix the comment when you fix the implementation.

DreadPirateRobertz (commented):
## 5-Agent Review — PR #241 (tl-uzn WebSocket Battle-State API)

Reviewers: code-reviewer · silent-failure-hunter · type-design-analyzer · comment-analyzer · pr-test-analyzer

---

### 🔴 BLOCKING — Must fix before merge

**1. Algorithm confusion attack — parseBattleWSToken missing HMAC assertion**

```go
func parseBattleWSToken(tokenStr, secret string) (*jwt.Token, error) {
    return jwt.Parse(tokenStr, func(t *jwt.Token) (any, error) {
        // MISSING: if _, ok := t.Method.(*jwt.SigningMethodHMAC); !ok { return nil, ... }
        return []byte(secret), nil
    })
}
```
An attacker can forge a token signed with RS256 (or `alg: none`) by providing their own public key as the "secret". Without the signing method assertion, the parser accepts it. This is a well-known jwt-go vulnerability class. **Add the guard:**
```go
if _, ok := t.Method.(*jwt.SigningMethodHMAC); !ok {
    return nil, fmt.Errorf("unexpected signing method: %v", t.Header["alg"])
}
```

**2. Missing test: TestBattleWS_AlgorithmConfusionTokenReturns401**

The algorithm confusion attack path has no test. Add a test that submits a token signed with RS256/none and asserts HTTP 401 is returned. This test must exist before merge to prevent future regressions from accidentally re-opening the vulnerability.

---

### 🟠 HIGH — Silent failures

**3. WriteJSON errors silently discarded in broadcast loop**

```go
if err := conn.WriteJSON(event); err != nil {
    // nothing — error dropped
}
```
Failed writes should disconnect the subscriber (close done channel) and log the error. A broken connection that never cleans up will hold a slot in the hub indefinitely.

**4. Ping write errors silently discarded**

```go
_ = conn.WriteMessage(websocket.PingMessage, nil)
```
A failed ping means the connection is dead. Treat ping failure the same as a failed WriteJSON: log + close done channel.

**5. NextReader errors: all errors treated as clean close**

```go
if _, _, err := conn.NextReader(); err != nil {
    break
}
```
A network error mid-message is not a clean close. Log the error (at debug level for `CloseError`, warn for others) before breaking.

**6. Hub event drops are invisible**

When a subscriber's channel is full (sendBuffer=16), events are silently dropped. Add a counter/log at drop time: `slog.WarnContext(ctx, "battle event dropped: subscriber buffer full", "battle_id", battleID, "subscriber_id", subID)`.

---

### 🟡 COMMENT ACCURACY

**7. wsUpgrader CheckOrigin comment is a false security claim — CRITICAL documentation bug**

```go
// Browsers enforce same-origin policy on WebSocket upgrades;
// the JWT in the query-string is the auth layer.
```
This is wrong. Browsers do NOT enforce same-origin policy on WebSocket upgrades — that's the entire premise of CSWSH (Cross-Site WebSocket Hijacking). The JWT in the query string IS the auth layer, but the comment's reasoning is inverted. Replace with:
```
// CheckOrigin is set to allow all origins. The bearer token in the
// query string provides authentication; ambient cookies are not used,
// so CSRF does not apply. For additional hardening, configure an
// allowlist from the environment.
```

**8. parseBattleWSToken godoc claims parity with middleware.Authenticate — no longer true after fix**

The docstring says "validates the token using the same logic as middleware.Authenticate." After adding the HMAC assertion (fix #1), ensure the docstring is updated to document the method constraint explicitly so future maintainers know the function is HMAC-only.

---

### 🟡 TYPE DESIGN

**9. `Subscriber.C` should be `<-chan Event` (receive-only)**

`C chan Event` allows callers outside the package to send to or close the channel, both of which corrupt hub state. Change to `C <-chan Event`. The hub holds the bidirectional reference internally; external callers only need to receive.

**10. `EventType` should be a newtype, not an untyped string constant**

`const EventDamage = "damage"` accepts any raw string in `Broadcast(battleID, eventType string, ...)`. Change:
```go
type EventType string
const EventDamage EventType = "damage"
// ...
func (h *Hub) Broadcast(battleID string, eventType EventType, payload interface{})
```

**11. `Event.Payload` as `interface{}` is not compile-time JSON-safe**

A non-marshallable payload (channel, func) will silently produce a broken JSON frame. Consider `json.RawMessage` — callers marshal before calling Broadcast, moving the error to the call site.

**12. Join frame has different wire shape than all other events**

The join frame is `map[string]string{"type":"joined","battle_id":...}` while all other events go through the `Event` envelope (with `timestamp`, `payload`). Clients can't deserialise all frames uniformly. Unify: send join as a proper `Event{Type: "joined", BattleID: battleID}`.

---

| Severity | Count |
|----------|-------|
| 🔴 BLOCKING | 2 |
| 🟠 HIGH | 4 |
| 🟡 Notable | 6 |

Items 1 + 2 (algorithm confusion + test) are security blocking. Items 3–6 (silent failures) should also be fixed before merge — a hub that accumulates dead connections will OOM in production. Items 7–8 (comment accuracy) fix a false security claim that will mislead future maintainers.

— Petra (5-agent: code-reviewer + silent-failure-hunter + type-design-analyzer + comment-analyzer + pr-test-analyzer)

DreadPirateRobertz:
Closing: the WebSocket feature (tl-uzn) landed on main via . The CodeQL log-injection fix (sanitizeForLog) was applied directly to main's battle_ws.go in . PR branch used a different Hub architecture (wsbattle package) that conflicted with the already-merged handler.Hub approach — security fix is now in main.

PR #258, opened 2026-04-17
feat(tl-5wz): spaced-repetition scheduler — SM-2 on concept slugs + injectable clock

DreadPirateRobertz:
## What
New spaced-repetition scheduler in `tutoring-service`, distinct from the existing per-course `/reviews/*` queue:

- `POST /v1/spaced-repetition/review-result` — record an SM-2 review response.
- `GET  /v1/spaced-repetition/due`           — list concepts currently due (overdue first).

Backed by a new `concept_review_schedule` table keyed on **global concept slugs** (text) rather than per-course `Concept` UUIDs. Reuses the existing pure-function `sm2_update()` in `app/srs.py`; adds `app/clock.py`, a `Callable[[], datetime]` FastAPI dependency so time can be frozen in tests.

## Why
Bead ID: **tl-5wz** — Phase 5: spaced-repetition scheduler, proactive review prompts on the global concept graph. This complements (doesn't replace) the `/reviews/*` queue on `StudentConceptMastery`, which is scoped to per-course concept UUIDs; this new scheduler operates on the subject-wide concept graph slugs that the tutoring loop emits.

### New / Changed
- `infra/db/migrations/019_concept_review_schedule.{up,down}.sql` — new table + RLS + `ix_crs_user_due` composite index.
- `app/orm.py` — `ConceptReviewSchedule` model.
- `app/models.py` — `SpacedRepetitionReview{Request,Response}` + `SpacedRepetitionDue{Item,Response}`.
- `app/clock.py` — `Clock` type alias + `get_clock()` FastAPI dependency (default `datetime.now(timezone.utc)`).
- `app/spaced_repetition.py` — two endpoints, both `Depends(get_clock)` and `Depends(require_auth)`.
- `app/main.py` — router registration under `/v1`.
- `tests/test_spaced_repetition.py` — 14 tests, **100% coverage** on `app/spaced_repetition.py` and `app/clock.py`.

## How to test
1. Apply migration `019_concept_review_schedule.up.sql` against a dev Postgres with the `users` table present.
2. `cd services/tutoring-service && pytest tests/test_spaced_repetition.py -v` — all 14 pass.
3. Auth a user, `POST /v1/spaced-repetition/review-result` with `{"concept_id": "chemistry.organic.stereochemistry", "quality": 5}`. Expect `interval_days=1`, `repetitions=1`, `ease_factor=2.6`, `due_at` one day later.
4. Repeat with `quality=4` — interval advances to 6 days (SM-2 second-success path).
5. Call `GET /v1/spaced-repetition/due` — the newly scheduled concept should **not** appear (due 1 day out); rows with `quality<3` or never-reviewed rows (`due_at IS NULL`) appear in the list.
6. Tests override `get_clock` via `app.dependency_overrides` to freeze time — confirms deterministic due-date computation.

## Checklist
- [x] Tests written (TDD) — `tests/test_spaced_repetition.py` (14 tests)
- [x] Coverage ≥90% on new code — **100%** on both new modules (see CI Codecov report)
- [x] Docstrings on all new functions/classes (Google-style)
- [x] Lint clean — `ruff check` passes
- [x] No secrets committed
- [x] Self-review: read diff before opening

### Current behaviour (before this PR)
Only `/v1/reviews/*` existed — operating on `student_concept_mastery` (per-course `Concept` UUID keys). There was no way for the tutoring loop to schedule reviews against the global concept graph's slug IDs, and no dependency-injected clock for deterministic time in tests.

### New behaviour (after this PR)
Two JWT-protected endpoints on `/v1/spaced-repetition/*` operate on text `concept_id` slugs. `get_clock` is overridable via `app.dependency_overrides` so integration tests can freeze and advance time. `sm2_update` is reused unchanged from `app/srs.py`.

### Detail
- `concept_review_schedule` primary key is `(user_id, concept_id)`; `concept_id` is `TEXT` (no FK — the slug namespace lives outside this service).
- Composite index `ix_crs_user_due (user_id, due_at)` powers the due-list query without a sort.
- Quality < 3 resets `repetitions=0` and `interval_days=1` (next day reappearance); ease factor retained.
- First passing review → interval 1 day; second → 6 days; thereafter `round(interval * ease_factor)` per SM-2.
- RLS policy `user_isolation` on `user_id = current_setting('app.current_user_id')` matches the existing `review_records` policy.

### DIFF
--- a/infra/db/migrations/019_concept_review_schedule.down.sql
+++ b/infra/db/migrations/019_concept_review_schedule.down.sql
@@ -1,0 +1,7 @@
+-- Rollback: drop concept_review_schedule table (tl-5wz)
+
+BEGIN;
+
+DROP TABLE IF EXISTS concept_review_schedule CASCADE;
+
+COMMIT;
--- a/infra/db/migrations/019_concept_review_schedule.up.sql
+++ b/infra/db/migrations/019_concept_review_schedule.up.sql
@@ -9,7 +9,7 @@ CREATE EXTENSION IF NOT EXISTS pgcrypto;
 CREATE TABLE concept_review_schedule (
   user_id          UUID        NOT NULL REFERENCES users(id) ON DELETE CASCADE,
   concept_id       TEXT        NOT NULL,
-  ease_factor      FLOAT       NOT NULL DEFAULT 2.5,
+  easiness_factor  FLOAT       NOT NULL DEFAULT 2.5,
   interval_days    INT         NOT NULL DEFAULT 1,
   repetitions      INT         NOT NULL DEFAULT 0,
   last_reviewed_at TIMESTAMPTZ,
--- a/infra/db/migrations/019_concept_review_schedule.up.sql
+++ b/infra/db/migrations/019_concept_review_schedule.up.sql
@@ -1,0 +1,28 @@
+-- TeachersLounge — concept review schedule (tl-5wz)
+-- Per-student SM-2 schedule keyed on global concept slugs (text concept_id).
+-- Distinct from student_concept_mastery, which is keyed on per-course Concept UUIDs.
+
+BEGIN;
+
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
+
+CREATE TABLE concept_review_schedule (
+  user_id          UUID        NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+  concept_id       TEXT        NOT NULL,
+  ease_factor      FLOAT       NOT NULL DEFAULT 2.5,
+  interval_days    INT         NOT NULL DEFAULT 1,
+  repetitions      INT         NOT NULL DEFAULT 0,
+  last_reviewed_at TIMESTAMPTZ,
+  due_at           TIMESTAMPTZ,
+  PRIMARY KEY (user_id, concept_id)
+);
+
+-- Powers GET /spaced-repetition/due: filter by user, sort by due_at ascending.
+CREATE INDEX ix_crs_user_due ON concept_review_schedule (user_id, due_at);
+
+ALTER TABLE concept_review_schedule ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY user_isolation ON concept_review_schedule
+    USING (user_id = current_setting('app.current_user_id', true)::uuid);
+
+COMMIT;
--- a/services/tutoring-service/app/chat.py
+++ b/services/tutoring-service/app/chat.py
@@ -50,6 +50,7 @@ from .models import MessageRequest
 from .rag import reformulate_query
 from .rag_agent import PROFESSOR_NOVA_SYSTEM_PROMPT, build_rag_context
 from .search_client import fetch_diagram_chunks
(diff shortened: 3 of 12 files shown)


DreadPirateRobertz:
## PR #258 Review — feat(tl-5wz): spaced-repetition scheduler

**2 blocking, 2 important. Changes requested before merge.**

---

### BLOCKING — 1: SM-2 EF not updated on failed recall

`srs.py`, `sm2_update` failure branch returns `ease_factor` unchanged:
```python
if quality < PASS_THRESHOLD:
    return 1, ease_factor, 0   # EF never decreases on failure
```
Canonical SM-2 applies the EF delta on every review. At `quality=1`: `delta = 0.1 - 4*(0.08 + 4*0.02) = -0.54` — EF should drop. As implemented, an item can be failed repeatedly without the easiness score ever reflecting that difficulty. The bug is also baked into the test (`test_failing_review_resets_repetitions_on_existing_row` asserts `ease_factor == pytest.approx(2.7)` unchanged).

**Fix:** apply the EF delta formula on the failure branch too, clamped to 1.3 minimum. Update the test to assert the reduced EF value.

---

### BLOCKING — 2: Session-start hook not wired

`sessions.py` has zero references to `spaced_repetition`, due-concept queries, or system-prompt injection. The spec deliverable — "inject due-concept reminder into system prompt when concepts are due" — is entirely absent from the diff. Router is registered in `main.py` but no session creation path calls the due-concepts check or augments the system prompt.

**Fix:** in the session-start path (wherever the system prompt is assembled), call `GET /spaced-repetition/due?user_id=X` (or the internal equivalent), and if concepts are due, prepend the reminder to the system prompt context.

---

### IMPORTANT — 3: `next_review_time` bare `datetime.now()` fallback

`srs.py` `next_review_time`: `base = now or datetime.now(timezone.utc)`. Clock injection contract says no bare `datetime.now()` in business logic. Make `now` a required parameter (no default) or remove the fallback.

---

### IMPORTANT — 4: Column name diverges from spec

Migration creates `ease_factor`; spec and task description say `easiness_factor`. Rename now before deploy — a post-deploy rename needs another migration.

---

### What's clean
Clock injection correctly threaded through `spaced_repetition.py`. Migration present with correct column types. SM-2 success-branch math and EF clamping correct. Test coverage solid on success path and EF bounds. Docstrings present with Google-style `Args:`.

— Petra (PM)

DreadPirateRobertz:
## PR #258 Re-review — fixes verified ✓

**Approved. CI green. Merging.**

Both blocking issues resolved:

1. **SM-2 EF on failure**: EF delta now applied and clamped to 1.3 on the failure branch. Tests updated to assert the reduced EF value.
2. **Session-start hook wired**: `get_session_start_reminder` called in session creation path, result injected into `system_prompt` with graceful fallback on error. Clean.

Clock injection via `get_clock()` is correctly threaded — no bare `datetime.now()` in business logic (the private `_utc_now` backing `get_clock` is the only call site, which is correct). Column rename to `easiness_factor` confirmed in migration.

— Petra (PM)

